PR #40, opened 2025-11-13
feat: add AdCPBaseModel with exclude_none serialization

bokelley:
## Summary
Adds `AdCPBaseModel` base class that ensures spec-compliant JSON serialization where optional fields with `None` values are omitted entirely rather than sent as `null`. Updates all generated models to use this base class.

## Changes
- Create `AdCPBaseModel` that defaults `exclude_none=True` in `model_dump()` and `model_dump_json()`
- Update code generator to use `AdCPBaseModel` instead of `BaseModel`
- Regenerate all 74 Pydantic models with new base class
- Export `AdCPBaseModel` from `adcp.types`

## Testing
All 180 tests pass, including new serialization behavior verification.

### DIFF
--- a/src/adcp/types/base.py
+++ b/src/adcp/types/base.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 """Base model for AdCP types with spec-compliant serialization."""
 
+from typing import Any
+
 from pydantic import BaseModel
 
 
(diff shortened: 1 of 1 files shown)


youbek (approved):
lgtm

PR #210, opened 2026-04-20
feat(schemas): regen for AdCP 3.0 GA — custom pricing + experimental_features

### LINKED ISSUE #204: Regen for AdCP 3.0 GA: new `custom` pricing variant + `experimental_features[]`
AdCP 3.0 GA ships three schema changes affecting generated Pydantic models. All additive to the wire format.

Upstream PR: adcontextprotocol/adcp#2422

## Changes to regenerate

1. **New `custom` variant in vendor pricing `Union[...]`.**
   ```python
   class CustomPricing(BaseModel):
       model: Literal["custom"]
       description: str       # required
       metadata: Dict[str, Any]   # required, non-empty
       currency: str | None = None  # ISO 4217 when monetary
   ```
   Consumers using `match p.model` or discriminated-union `isinstance` checks need to add a `CustomPricing` branch. Recommended default: route to operator review rather than auto-select.

2. **New optional `experimental_features: list[str] | None` on `GetAdcpCapabilitiesResponse`.**
   Dot-namespaced feature ids that sellers declare when they implement experimental surfaces (e.g. `brand.rights_lifecycle`, `governance.campaign`, `trusted_match.core`). Consumers check membership before relying on experimental surfaces. Full contract: https://adcontextprotocol.org/docs/reference/experimental-status

3. **`x-status: experimental` annotations on Brand Rights, Campaign Governance, and TMP schemas.**
   Metadata only — `datamodel-code-generator` and most Pydantic generators ignore unknown `x-*` extensions. Type output is unchanged. Optionally surface in docstrings.

## Pre-existing backlog

If `per_unit` pricing variant (added in an earlier 3.0 rc) is not yet in your generated models, this regen will pick it up alongside `custom`.

## Suggested release

- Minor version bump
- Changelog: "Handle `model: 'custom'` in pricing consumers. New `experimental_features` attribute on capabilities."
- No wire-format breaks

bokelley:
Closes #204. Regenerates Pydantic models from the upstream AdCP 3.0 GA protocol bundle ([adcontextprotocol/adcp#2422](https://github.com/adcontextprotocol/adcp/pull/2422)).

## Summary

- New `custom` variant in vendor pricing discriminated union (`VendorPricing`, `VendorPricingOption`). Requires `description` and `metadata` (non-empty), optional `currency`. Consumers dispatching on `.root.model` should add a `"custom"` branch — recommended default is operator review, not auto-select.
- `per_unit` pricing variant (added in an earlier 3.0 rc) is now in the generated models alongside `custom`.
- New optional `experimental_features: list[str] | None` on `GetAdcpCapabilitiesResponse`. Dot-namespaced ids (`brand.rights_lifecycle`, `governance.campaign`, `trusted_match.core`) declared by sellers that implement experimental surfaces.
- `x-status: experimental` annotations on Brand Rights, Campaign Governance, and TMP — metadata only; `datamodel-code-generator` ignores unknown `x-*` extensions, so type output is unchanged.
- `Idempotency` now carries a required `supported: bool` (pulled in via the latest bundle). Four test fixtures updated to pass `supported=True`.

## Wire format

No breaks. All changes are additive.

## Test plan

- [x] `make regenerate-schemas` succeeds
- [x] `ruff check src/` passes
- [x] `mypy src/adcp/` passes (671 source files)
- [x] `pytest tests/` passes (1696 passed, 14 skipped)
- [x] Verified `VendorPricing5`/`VendorPricingOption11` carry `model: Literal["custom"]` with required `description` + `metadata`
- [x] Verified `experimental_features` field on `GetAdcpCapabilitiesResponse` in both `protocol/` and `bundled/protocol/`


bokelley:
## Expert review follow-up

Ran the PR past three expert reviewers. Summary of findings and actions:

### Addressed in 9ff08c9b (follow-up commit)
**[must-fix from code-reviewer]** `_ensure_idempotency_capability` in `src/adcp/client.py` ignored the new required `supported` field. A seller declaring `supported=false, replay_ttl_seconds=86400` (contradictory but type-valid) passed the strict gate silently. Fixed to raise `IdempotencyUnsupportedError` on any of:
- `adcp.idempotency` missing entirely
- `supported=false` (seller does not dedupe — naive retry double-processes)
- `supported=true` but `replay_ttl_seconds is None` (seller is spec-incomplete)

Updated `IdempotencyUnsupportedError` to carry a `reason` string so logs show which mode tripped. Added three tests covering each failure mode plus the happy path.

### Not addressed (acknowledged, scoped out of this PR)

- **`Metadata.minProperties: 1` not enforced on custom pricing** (ad-tech-protocol-expert, python-expert). The JSON Schema says the `metadata` object on custom-pricing variants must be non-empty, but `datamodel-codegen` flattens `additionalProperties: true` + optional `summary_for_operator` to a model that accepts `{}`. This is a codegen limitation, not something this regen introduced. Worth a follow-up as a `post_generate_fixes.py` hook or a hand-written `model_validator` that survives regen.
- **`Idempotency` conditional `if supported then require ttl` not enforced at the model level** (ad-tech-protocol-expert). Same class of codegen gap (draft-07 `if/then` isn't emitted). The client-side gate fix above compensates for the buyer path; a seller building capabilities responses should still own spec conformance.
- **`VendorPricing` / `VendorPricingOption` / `ExperimentalFeature` / capability-response `Idempotency` not re-exported via `adcp.types` public API** (python-expert). Tests in `test_capabilities.py` reach into `adcp.types.generated_poc.protocol.get_adcp_capabilities_response` directly — this pre-existed this PR. Adding public re-exports + semantic aliases (`CustomPricing`, etc.) is worth doing but is a distinct DX improvement, not a regen correctness fix.

### Confirmed correct / no action needed
- Custom-variant fields match spec: `model`, `description` (min_length=1), `metadata` required; `currency` (ISO-4217 pattern) and `ext` optional. `extra="allow"` preserves escape-hatch semantics for vendor-specific metadata keys.
- `experimental_features` field correctly propagates the dot-namespace pattern and is optional (matches wire contract — seller absence = "implements none").
- Making `Idempotency.supported` required matches the spec posture ("clients MUST NOT assume a default").
- Multiple-inheritance subclasses on `VendorPricingOption7..11` dispatch deterministically via the `model` Literal discriminant under Pydantic v2. No MRO pitfall.
- No SDK-internal dispatch on pricing `.model` exists (verified via grep) — consumers are responsible for adding a `custom` branch per the issue's guidance.
- `aliases.py` intentionally omits per-variant signal-pricing aliases (see existing comment at lines 1062-1067); adding `CustomPricing` here alone would be inconsistent with the established pattern.

PR #251, opened 2026-04-21
feat(a2a): auto-retain contextId + taskId across multi-turn calls

bokelley:
## Summary

- Fixes a regression introduced in v3.0.0-rc.2 where the A2A adapter dropped `Message.context_id` entirely (v2.12.0 minted a fresh UUID per send; v3 sent nothing). Both broke multi-turn buyer flows that need server-side session continuity.
- Adds `task_id` retention for non-terminal states (`input-required`, `working`, `submitted`, `auth-required`) so HITL / approval flows can resume the same server-side task instead of orphaning it.
- Covers the resume-across-process-restart story: callers can persist `client.context_id` and rehydrate a new `ADCPClient` with `context_id=<saved>` to land on the same session.

### Public API (A2A only)
```python
ADCPClient(config, context_id="resume-or-self-named")  # seed
client.context_id                                      # read
client.pending_task_id                                 # read (resume state)
client.reset_context()                                 # clear — start new convo
client.reset_context("new-id")                         # clear + seed
```

Non-A2A callers: reads return `None` (safe in generic code); writes raise `ValueError` — symmetric writes, lenient reads, explicitly documented in the property docstring.

### Design notes

- **Default on** — auto-retain is the pit of success for multi-turn. Docstring calls out the "one ADCPClient per conversation" rule and the last-write-wins race if shared across concurrent conversations.
- **ADK-aware** — the ADK server rewrites buyer-proposed UUIDs into its own `ADK/<app>/<user>/<session>` format and returns the rewritten id. The adapter auto-adopts whatever the server returns, so buyer-seeded ids survive rewriting transparently.
- **task_id lifecycle** — retained only while the last response was non-terminal; cleared on terminal states so subsequent calls start fresh tasks under the same context.

### Test coverage
- 24 new unit tests in `tests/test_protocols.py` covering auto-retain, inject, reset, server rebinding, task_id retention in each of the non-terminal states, task_id clearing on terminal states, and non-A2A rejection.
- 5 new HTTP integration tests in `tests/integration/test_a2a_context_id.py` that spin up a real a2a-sdk Starlette app under uvicorn and prove the contract end-to-end, including the HITL `input-required` → resume path.

## Test plan

- [x] `ruff check` clean
- [x] `mypy src/adcp/` clean
- [x] `pytest tests/test_protocols.py tests/integration/test_a2a_context_id.py` — 53 passed
- [x] Full suite `pytest tests/` — 2022 passed (excluding two pre-broken files missing `asgi_lifespan`, unrelated)
- [ ] Manual smoke against `test-agent.adcontextprotocol.org` (blocked: that agent's `/.well-known/agent-card.json` returns 404, also breaks the existing reference-agent A2A tests — separate issue)

### DIFF
--- a/src/adcp/client.py
+++ b/src/adcp/client.py
@@ -309,6 +309,7 @@ class ADCPClient:
         validate_features: bool = False,
         strict_idempotency: bool = False,
         signing: SigningConfig | None = None,
+        context_id: str | None = None,
     ):
         """
         Initialize ADCP client for a single agent.
@@ -343,6 +344,20 @@ class ADCPClient:
                 ``jwks_uri``. Supported on both A2A and MCP
                 (``mcp_transport="streamable_http"``); SSE-transport MCP
                 logs a warning and falls through unsigned.
+            context_id: A2A-only. Seed the A2A conversation context. Pass a
+                previously-returned ``context_id`` to resume a session
+                across process restarts, or a self-assigned UUID to name
+                the session with your own correlation key (the ADK server
+                honors buyer-proposed ids). If omitted, the server mints
+                one on the first message and this client auto-retains it
+                for subsequent calls. Read the current value via
+                ``client.context_id``; call ``client.reset_context()`` to
+                start a fresh conversation. Rule of thumb: one
+                ``ADCPClient`` per A2A conversation — if a buyer has
+                multiple concurrent briefs with the same agent, construct
+                one client per brief rather than sharing.
+
+                Raises ``ValueError`` if passed with a non-A2A protocol.
         """
         self.agent_config = agent_config
         self.webhook_url_template = webhook_url_template
@@ -380,10 +395,80 @@ class ADCPClient:
         if signing is not None:
             self.adapter.signing_request_hook = self._sign_outgoing_request
 
+        if context_id is not None:
+            if not isinstance(self.adapter, A2AAdapter):
+                raise ValueError(
+                    "context_id is only supported for A2A protocol; " f"got {agent_config.protocol}"
+                )
+            self.adapter.set_context_id(context_id)
+
         # Initialize simple API accessor (lazy import to avoid circular dependency)
         from adcp.simple import SimpleAPI
 
         self.simple = SimpleAPI(self)
+
+    @property
+    def context_id(self) -> str | None:
+        """Current A2A conversation context_id.
+
+        Reads the context_id currently associated with this client: the
+        value assigned by the A2A server (auto-captured from the most
+        recent response) or the one seeded via the constructor or
+        ``reset_context()``. Returns ``None`` before the first A2A call
+        in a fresh conversation, or for clients on non-A2A protocols —
+        reads are lenient across protocols so generic code can probe
+        ``if client.context_id: ...`` safely. Writes (constructor kwarg,
+        ``reset_context``) raise on non-A2A because the operation has no
+        meaning there.
+
+        Not safe for concurrent calls on the same client — the adapter
+        mutates this on every response. Rule of thumb: one ADCPClient
+        per A2A conversation. Persist this value (e.g., Redis keyed by
+        your brief id) to resume across process restarts by passing it
+        to ``ADCPClient(context_id=...)``.
+        """
+        if isinstance(self.adapter, A2AAdapter):
+            return self.adapter.context_id
+        return None
+
+    @property
+    def pending_task_id(self) -> str | None:
+        """A2A task_id pending resume, or None if no task is in-flight.
+
+        Set when the last A2A response was non-terminal
+        (``input-required``, ``working``, ``submitted``,
+        ``auth-required``). The adapter echoes this id on the next
+        outbound message so the server resumes the same task. Clears
+        automatically when the task reaches a terminal state.
+
+        Returns ``None`` for non-A2A clients.
+        """
+        if isinstance(self.adapter, A2AAdapter):
+            return self.adapter.pending_task_id
+        return None
+
+    def reset_context(self, context_id: str | None = None) -> None:
+        """Start a new A2A conversation on this client.
+
+        Passing ``None`` (default) clears the current context so the
+        server mints a fresh one on the next call. Passing a string uses
+        it as the new conversation id — useful for resuming a specific
+        prior session or for naming the conversation with your own
+        correlation key. Note: some servers (notably ADK) rewrite
+        client-supplied ids into their own session format; the client
+        auto-adopts the rewritten id on the next response.
+
+        Also clears any pending_task_id — starting a new conversation
+        discards any in-flight task on the old one.
+
+        Raises ``ValueError`` when called on a non-A2A client.
+        """
+        if not isinstance(self.adapter, A2AAdapter):
+            raise ValueError(
+                "reset_context is only supported for A2A protocol; "
+                f"got {self.agent_config.protocol}"
+            )
+        self.adapter.set_context_id(context_id)
 
     async def _ensure_idempotency_capability(self) -> None:
         """Verify the seller positively declares idempotency support in capabilities.
--- a/src/adcp/protocols/a2a.py
+++ b/src/adcp/protocols/a2a.py
@@ -38,11 +38,81 @@ logger = logging.getLogger(__name__)
 class A2AAdapter(ProtocolAdapter):
     """Adapter for A2A protocol using official a2a-sdk client."""
 
+    # A2A task states in which the server is still expecting more from
+    # the buyer on the same task (input-required, auth-required, and
+    # in-flight states). While the adapter holds a task_id in one of
+    # these states, the next outbound Message must echo it back so the
+    # server resumes the same task rather than orphaning it and starting
+    # a new one. Terminal states (completed/failed/canceled/rejected)
+    # clear the retained task_id — subsequent calls in the conversation
+    # are new tasks.
+    _NONTERMINAL_TASK_STATES = frozenset(
+        {"submitted", "working", "input-required", "auth-required"}
+    )
+
     def __init__(self, agent_config: AgentConfig):
         """Initialize A2A adapter with official A2A client."""
         super().__init__(agent_config)
         self._httpx_client: httpx.AsyncClient | None = None
         self._a2a_client: A2AClient | None = None
+        # A2A contextId for multi-turn conversations. First request sends
+        # context_id=None → server mints one and returns it on Task.context_id;
+        # we stash it here and echo it back on every subsequent send so the
+        # server can scope state to the same session. Callers can seed this
+        # via ADCPClient(context_id=...) to resume a session across process
+        # restarts, or clear it via ADCPClient.reset_context() to start a
+        # new conversation.
+        self._context_id: str | None = None
+        # A2A task_id retained across turns only while the prior task is
+        # non-terminal (input-required, working, etc). On terminal states
+        # this clears to None so the next call starts a new task under
+        # the same context_id. Without this, resume of an input-required
+        # task orphans the server-side pending task.
+        self._pending_task_id: str | None = None
+
+    @property
+    def context_id(self) -> str | None:
+        """Current A2A conversation context_id, or None if not yet established.
+
+        ``None`` means either (a) a fresh conversation where the server
+        has not yet replied, or (b) the context was cleared via
+        ``set_context_id(None)``. Callers that need to distinguish these
+        must track their own state.
+
+        Not thread-safe: the adapter mutates this on every response. For
+        concurrent use, serialize calls on one adapter or construct one
+        per conversation.
+        """
+        return self._context_id
+
+    @property
+    def pending_task_id(self) -> str | None:
+        """A2A task_id retained for resume, or None if no task is pending.
+
+        Populated when the last response was non-terminal (e.g.
+        ``input-required``). Echoed on the next outbound message so the
+        server continues the same task. Clears to None on terminal
+        states (``completed``/``failed``/``canceled``).
+        """
+        return self._pending_task_id
+
+    def set_context_id(self, context_id: str | None) -> None:
+        """Set the A2A context_id for subsequent message sends.
+
+        Pass ``None`` to clear — the server mints a fresh id on the next
+        call — or a string to seed. Seeding is safe for *resume* (pass
+        back an id the server previously returned). Seeding with a
+        *self-generated* id is server-dependent: per the A2A spec,
+        agents MAY accept or reject client-supplied ids, and some
+        frameworks (notably ADK) rewrite the id into their own session
+        format and return the rewritten value on the next response — at
+        which point this adapter auto-adopts it.
+
+        Also clears any retained ``pending_task_id``: switching context
+        always starts a fresh task under the new context.
+        """
+        self._context_id = context_id
+        self._pending_task_id = None
 
     async def _get_httpx_client(self) -> httpx.AsyncClient:
         """Get or create the HTTP client with connection pooling."""
@@ -189,6 +259,8 @@ class A2AAdapter(ProtocolAdapter):
                 message_id=message_id,
                 role=Role.user,
                 parts=[Part(root=data_part)],
+                context_id=self._context_id,
+                task_id=self._pending_task_id,
             )
         else:
             # Natural language invocation (flexible)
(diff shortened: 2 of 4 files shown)


src/adcp/protocols/a2a.py
```diff
@@ -265,6 +339,18 @@ async def _call_a2a_tool(
 
                 # Result can be either Task or Message
                 if isinstance(result, Task):
+                    # Retain the server-assigned context_id so subsequent
+                    # turns continue the same A2A conversation. Task.context_id
+                    # is required by a2a-sdk, so no None-guard needed.
+                    self._context_id = result.context_id
```
scr-oath, line 345: issue (blocking): state is mutated before a call that can raise, leaving the adapter in a state that reflects a response the caller never received.

`self._context_id` and `self._pending_task_id` are updated at lines 345–353 *before* `self._process_task_response(result, debug_info)` at line 354. If `_process_task_response` raises (malformed artifact, missing required field, etc.), the caller sees the exception but the adapter has already advanced its state. A retry would then send the wrong `task_id` — possibly `None` after a clear, or a new server-assigned id that the prior-turn server doesn't know about — potentially orphaning the legitimate in-flight task.

Fix: capture locals first, commit only after the processing step succeeds.

```python
new_context_id = result.context_id
new_pending_task_id = (
    result.id if result.status.state in self._NONTERMINAL_TASK_STATES else None
)
task_result = self._process_task_response(result, debug_info)  # may raise
_idempotency.raise_for_idempotency_error(tool_name, task_result.data, self.agent_config.id)  # may raise
# commit after all failure points
self._context_id = new_context_id
self._pending_task_id = new_pending_task_id
return _idempotency.annotate_result(task_result, idempotency_key)
```

src/adcp/protocols/a2a.py
```diff
@@ -265,6 +339,18 @@ async def _call_a2a_tool(
...
+                    # Retain task_id only while the task is non-terminal.
+                    # On terminal states (completed/failed/canceled/rejected)
+                    # the next send must NOT echo this task_id — it starts a
+                    # fresh task under the same context.
+                    if result.status.state in self._NONTERMINAL_TASK_STATES:
+                        self._pending_task_id = result.id
+                    else:
+                        self._pending_task_id = None
```
scr-oath, line 353: issue: the state-retention logic only runs inside `if isinstance(result, Task):`. The `else` branch (when `send_message` returns a bare `Message`, a few lines below) leaves both `_context_id` and `_pending_task_id` unchanged, with no test locking that decision down.

If a server returns a bare `Message` while the adapter holds a `_pending_task_id` from a prior `input-required`, the next outbound call will still echo that stale `task_id` — either correct (task still in-flight) or a bug (server ended via Message). In practice this branch may be unreachable against ADCP-conformant servers (per `a2a-response-format.mdx`, Task is always returned), but please either (a) defensively clear `_pending_task_id` in the Message branch with a rationale comment, or (b) add an explicit test that constructs `SendMessageSuccessResponse(result=<Message>)` and asserts the chosen behavior so future refactors don't silently change it.

Also worth confirming: does `a2a.types.Message` carry `context_id`? The PR title references "dropped `Message.context_id`" — if that field exists on Message, the fix is incomplete (capture from the Message result too). A one-line clarifying comment would help future readers either way.

tests/test_protocols.py
```diff
@@ -442,6 +442,323 @@ async def test_get_agent_info_without_extensions(self, a2a_config):
...
+            return_value=SendMessageSuccessResponse(result=failed)
+        )
+
+        with patch.object(adapter, "_get_a2a_client", return_value=mock_a2a_client):
+            await adapter._call_a2a_tool("get_products", {})
+
+        assert adapter.pending_task_id is None
+
+    @pytest.mark.asyncio
+    async def test_task_id_retained_on_working_state(self, a2a_config):
```
scr-oath, line 641: issue: `_NONTERMINAL_TASK_STATES` has four members — `submitted`, `working`, `input-required`, `auth-required` — but only `input-required` and `working` have unit tests asserting `pending_task_id` retention. `submitted` and `auth-required` have no equivalent. If a future refactor accidentally drops either from the frozenset, no unit test catches the regression (the integration tests only exercise `input-required` via `_HitlExecutor`).

Add mirror tests:

```python
async def test_task_id_retained_on_submitted_state(self, a2a_config): ...
async def test_task_id_retained_on_auth_required_state(self, a2a_config): ...
```

The PR description claims "task_id retention in each of the non-terminal states" — currently only half the states are covered.

tests/test_protocols.py
```diff
@@ -442,6 +442,323 @@ async def test_get_agent_info_without_extensions(self, a2a_config):
...
+            assert adapter.pending_task_id is None
+
+            await adapter._call_a2a_tool("create_media_buy", {})
+
+        assert self._captured_task_id(mock_a2a_client.send_message, 1) is None
+        second_call = mock_a2a_client.send_message.call_args_list[1]
+        assert second_call[0][0].params.message.context_id == "ctx-session"
+
+    @pytest.mark.asyncio
+    async def test_task_id_cleared_on_failed_state(self, a2a_config):
```
scr-oath, line 620: suggestion (non-blocking): `completed` and `failed` are tested for `pending_task_id` clearing, but `canceled` and `rejected` (both valid `TaskState` members with similar terminal semantics) are not. The logic is "anything not in the frozenset clears," so one test for either would lock the contract without ambiguity.

src/adcp/protocols/a2a.py
```diff
@@ -38,11 +38,81 @@
...
 
+    # A2A task states in which the server is still expecting more from
+    # the buyer on the same task (input-required, auth-required, and
+    # in-flight states). While the adapter holds a task_id in one of
+    # these states, the next outbound Message must echo it back so the
+    # server resumes the same task rather than orphaning it and starting
+    # a new one. Terminal states (completed/failed/canceled/rejected)
+    # clear the retained task_id — subsequent calls in the conversation
+    # are new tasks.
+    _NONTERMINAL_TASK_STATES = frozenset(
```
scr-oath, line 49: suggestion: `TaskState` has a 9th member — `unknown = 'unknown'` — that is implicitly treated as terminal here (clears `_pending_task_id`). Arguably correct (don't cling to an undefined-state task), but the class-level comment only enumerates `completed/failed/canceled/rejected` as terminal; `unknown` is silently lumped in. Worth adding either a one-line note to the comment or a `logger.warning` when encountered so operators notice if a server starts emitting it.

scr-oath, line 49: thought: consider coupling the frozenset to the `TaskState` enum directly:

```python
_NONTERMINAL_TASK_STATES: frozenset[TaskState] = frozenset({
    TaskState.submitted,
    TaskState.working,
    TaskState.input_required,
    TaskState.auth_required,
})
```

Gets IDE completion, type-checker coverage if a state is renamed upstream, and removes the implicit "is TaskState a str-Enum?" dependency. Trade-off is an explicit import coupling to `a2a.types.TaskState`, which you already have elsewhere. Also adds the missing `frozenset[str]` / `frozenset[TaskState]` type annotation on the class attribute.

src/adcp/protocols/a2a.py
```diff
@@ -38,11 +38,81 @@
...
+        """A2A task_id retained for resume, or None if no task is pending.
+
+        Populated when the last response was non-terminal (e.g.
+        ``input-required``). Echoed on the next outbound message so the
+        server continues the same task. Clears to None on terminal
+        states (``completed``/``failed``/``canceled``).
+        """
+        return self._pending_task_id
+
+    def set_context_id(self, context_id: str | None) -> None:
```
scr-oath, line 99: nitpick: `set_context_id` is exposed at the adapter level as a public method, but the intended caller-facing surface is `ADCPClient.reset_context()`. Consider renaming to `_set_context_id` so the public API funnels through `ADCPClient` — that also keeps the invariant (clearing `_pending_task_id` atomically with context changes) inescapable.

Current tests reach into `adapter.set_context_id(...)` directly; renaming would surface those as intentional internal-API uses.

src/adcp/client.py
```diff
@@ -380,11 +395,81 @@ def __init__(
         if signing is not None:
             self.adapter.signing_request_hook = self._sign_outgoing_request
 
+        if context_id is not None:
```
scr-oath, line 398: suggestion (non-blocking): the guard is `if context_id is not None`, but `context_id=""` slips through and sets `_context_id = ""`. An empty context on the wire is either a server error or a silently degenerate session. Tighten to `if context_id` (or add an explicit `and context_id != ""`) so callers using `context_id=os.getenv("...") or ""`-style patterns don't silently seed an empty id.

src/adcp/client.py
```diff
@@ -380,11 +395,81 @@ def __init__(
...
+            if not isinstance(self.adapter, A2AAdapter):
+                raise ValueError(
+                    "context_id is only supported for A2A protocol; " f"got {agent_config.protocol}"
```
scr-oath, line 401: suggestion: two minor points on this line.

**1. Mixed string concatenation** — adjacent string + f-string triggers ruff `ISC001`. Collapse to a single f-string:

```python
raise ValueError(
    f"context_id is only supported for A2A protocol; got {agent_config.protocol}"
)
```

**2. Exception type** — `ValueError` signals "bad value for this argument," but this is really "operation not supported on this protocol kind" (the string value is fine; the configuration is wrong for the operation). If the SDK already defines a protocol/operation exception (e.g. `ADCPError` subclass), prefer it here and at line 468 — callers who `except ADCPError:` will currently miss these. If no domain exception exists, `TypeError` is marginally closer to the intent ("wrong kind of object for this operation"), though either is defensible.

src/adcp/client.py
```diff
@@ -380,11 +395,81 @@ def __init__(
...
+        per A2A conversation. Persist this value (e.g., Redis keyed by
+        your brief id) to resume across process restarts by passing it
+        to ``ADCPClient(context_id=...)``.
+        """
+        if isinstance(self.adapter, A2AAdapter):
+            return self.adapter.context_id
+        return None
+
+    @property
+    def pending_task_id(self) -> str | None:
```
scr-oath, line 435: suggestion: naming — `pending_task_id` reads as "queued, not yet sent" to many API readers. The actual semantic is "a server-side task in a non-terminal state that the next call must echo to resume." `active_task_id` or `current_task_id` would convey ongoing work more clearly. Worth getting right before v3 GA since this lands in IDE autocomplete, logs, and persistence keys.

scr-oath (commented):
Post-merge audit review.

**Bottom line:** core logic is sound. `TaskState(str, Enum)` makes the `_NONTERMINAL_TASK_STATES` string-frozenset membership check safe, and the "retain task_id only on non-terminal, clear on terminal" rule matches A2A's resume-same-task contract. No `FunctionResponse` handling is needed from the client side — that is an ADK-internal concept; at the A2A protocol layer the client just sends `Message` and the ADK server converts internally.

**Risk level:** medium. Mostly additive; one real correctness concern (state mutated before a post-processing step that can raise) and two test-coverage gaps worth closing in a follow-up.

**Specialists consulted:** code-reviewer, python-pro, api-designer, test-automator.

**Counts:** 1 blocking issue + 2 test-coverage issues, 8 suggestions, several praises. Details inline.

**Notes on the enum-vs-string question** (`result.status.state in _NONTERMINAL_TASK_STATES`): verified against `a2a-sdk 0.3.24/25` — `class TaskState(str, Enum)` with values exactly `'submitted' | 'working' | 'input-required' | 'auth-required' | ...`, so enum members compare equal to their string values in set membership. Pydantic coerces string inputs to enum members in `A2ATaskStatus(state=...)`, so unit-test mocks exercise the real enum semantics, not a string-only happy path. Initial concern confirmed safe.

**ADCP metadata question**: the ADCP `context` echo field rides inside tool-call arguments (DataPart payload), not at the A2A Message envelope — so the adapter correctly doesn't touch it. Message-envelope fields `metadata`, `reference_task_ids`, `extensions` are not passed today; probably fine for the current scope, worth a tracking issue if buyers ever need message-level trace headers.

src/adcp/client.py
```diff
@@ -380,11 +395,81 @@ def __init__(
...
+        outbound message so the server resumes the same task. Clears
+        automatically when the task reaches a terminal state.
+
+        Returns ``None`` for non-A2A clients.
+        """
+        if isinstance(self.adapter, A2AAdapter):
+            return self.adapter.pending_task_id
+        return None
+
+    def reset_context(self, context_id: str | None = None) -> None:
```
scr-oath, line 450: suggestion: the docstring on the `context_id` property advertises the persist-and-resume story ("persist this value (e.g., Redis keyed by your brief id) to resume across process restarts"), but full resume state is *two* fields — `context_id` AND `pending_task_id`. A caller who only persists `context_id` and restores after an `input-required` response will reconnect to the right conversation but with `pending_task_id=None`, orphaning the in-flight task server-side.

Consider a first-class checkpoint/restore surface:

```python
def checkpoint(self) -> dict[str, str | None]:
    return {"context_id": self.context_id, "pending_task_id": self.pending_task_id}

@classmethod
def from_checkpoint(cls, agent_config, state, **kwargs):
    client = cls(agent_config, context_id=state.get("context_id"), **kwargs)
    if state.get("pending_task_id") and isinstance(client.adapter, A2AAdapter):
        client.adapter._pending_task_id = state["pending_task_id"]
    return client
```

Without this, the advertised Redis-resume pattern is subtly broken for HITL workflows — which is the primary use case the PR is motivated by.

tests/test_protocols.py
```diff
@@ -442,6 +442,323 @@ async def test_get_agent_info_without_extensions(self, a2a_config):
...
+            return_value=SendMessageSuccessResponse(result=working)
+        )
+
+        with patch.object(adapter, "_get_a2a_client", return_value=mock_a2a_client):
+            await adapter._call_a2a_tool("create_media_buy", {})
+
+        assert adapter.pending_task_id == "task-in-progress"
+
+    @pytest.mark.asyncio
+    async def test_set_context_id_clears_pending_task(self, a2a_config):
```
scr-oath, line 663: nitpick: this test and `test_clearing_context_id_starts_fresh_conversation` write directly to `adapter._context_id` / `adapter._pending_task_id` for setup, bypassing the public `set_context_id()` path. Using the public API for setup exercises the full write path and decouples tests from attribute names. If the direct write is intentional (isolating a specific code path), a comment noting that would help. `test_pending_task_id_property_exposes_adapter_state` has the same pattern.

src/adcp/protocols/a2a.py
scr-oath, line 345: praise: the `_NONTERMINAL_TASK_STATES` frozenset + the "retain only on non-terminal" rule is the right model for A2A's resume-same-task contract. The inline comment at lines 346–349 justifying *why* task_id must be cleared on terminal states is exactly the kind of breadcrumb that prevents future regressions from "just retain it always, simpler that way" refactors. Confirmed safe against `a2a-sdk 0.3.24/25` where `TaskState` is `class TaskState(str, Enum)`.

src/adcp/protocols/a2a.py
```diff
@@ -38,11 +38,81 @@
...
+        agents MAY accept or reject client-supplied ids, and some
+        frameworks (notably ADK) rewrite the id into their own session
+        format and return the rewritten value on the next response — at
+        which point this adapter auto-adopts it.
+
+        Also clears any retained ``pending_task_id``: switching context
+        always starts a fresh task under the new context.
+        """
+        self._context_id = context_id
+        self._pending_task_id = None
```
scr-oath, line 115: praise: `set_context_id` atomically clearing `_pending_task_id` on every context change is the correct invariant. A "reset context but keep stale task_id" state would be a hard-to-diagnose bug class, and this method makes it unreachable by construction.

tests/integration/test_a2a_context_id.py
```diff
@@ -0,0 +1,403 @@
...
+        return None, {}
+
+
+def _pick_free_port() -> int:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("127.0.0.1", 0))
+        return int(s.getsockname()[1])
+
+
+@asynccontextmanager
```
scr-oath, line 120: praise: the module docstring honestly explains what server-side observation can and cannot prove (server always mints an id, so "client sent None" and "client sent X" look the same server-side on the first turn), and the tests then check the right observable: session continuity across turns. `_pick_free_port` via `bind(("127.0.0.1", 0))` is the right pattern for parallel-safe tests.

suggestion (non-blocking): small TOCTOU window between `s.close()` and uvicorn binding to that port. Under high-concurrency CI, another process can claim it. `uvicorn.Server.serve(sockets=[s])` accepts a pre-bound socket directly, avoiding the close-then-rebind gap. Very low priority.

PR #560, opened 2026-05-04
fix(server): echo request context on AdcpError error envelopes

### LINKED ISSUE #557: bug(server): AdcpError raise path bypasses context-echo on MCP and A2A error envelopes
## Bug

The MCP and A2A error paths build the structured `adcp_error` envelope but never echo the request's `context` extension into the response. The success paths do echo it (`inject_context` in `mcp_tools.py:2030`), so any handler that raises `AdcpError` produces a response that violates the AdCP context-echo contract — buyers lose correlation IDs, idempotency hints, and other agent-to-agent state.

## Repro

A handler calls `raise AdcpError('PERMISSION_DENIED', message='nope')`. The wire response carries `structuredContent.adcp_error` but no top-level `context` field, even though the request supplied one.

## Affected paths

- **MCP**: `src/adcp/server/serve.py:1787-1804` — `except ADCPError` and the decisioning-AdcpError branch both call `build_mcp_error_result(exc)` without passing `kwargs` (the raw params).
- **A2A**: `src/adcp/server/a2a_server.py:237` — `_send_adcp_error` is invoked with `(event_queue, context, exc)` but not the parsed request `params` that carry the wire `context` extension.
- **Builders**: `src/adcp/server/translate.py:167` (`build_mcp_error_result`) and `src/adcp/server/a2a_server.py:426` (`_send_adcp_error`) take `exc` only, with no surface to receive raw request params.

## Symptom

Three storyboard runners fail on the AdcpError-raise path because the asserted echoed-context fields are missing from error responses.

## Fix

1. Plumb raw params into the error-result builders on both transports.
2. After building the structured envelope, run `inject_context(raw_params, envelope)` so the `context` field is echoed at the same level the success path uses.
3. Add tests covering both transports' error path with and without a request `context` field.

Plan to land as a single PR.

bokelley:
Closes #557

## Summary

Fixes a context-passthrough gap on the AdcpError raise path. The success path runs `inject_context(raw_params, response)` so a request's `context` extension echoes back to the buyer. The error path built the structured `adcp_error` envelope without doing the same — buyers lost correlation IDs and idempotency hints across the raise-AdcpError boundary, and three storyboard runners were failing on this assertion.

## What changed

- **`build_mcp_error_result(exc, params=None)`** now accepts the raw request dict. When supplied, it runs `inject_context` on the `structuredContent` dict so `context` lands as a sibling of `adcp_error`, mirroring the success path's wire shape.
- **MCP catch sites** (`src/adcp/server/serve.py`) pass `kwargs` through on both branches — the `ADCPError` catch and the decisioning-`AdcpError` re-projection.
- **`ADCPAgentExecutor._send_adcp_error(...)`** now accepts `params` and runs `inject_context` on the DataPart payload before sending the failed task.
- **`execute()`** in the A2A executor passes the parsed `params` into `_send_adcp_error`.

The change is backwards-compatible: both new params are optional. Existing callers (proxies, custom transports invoking `build_mcp_error_result` directly) keep working without context echo until they wire it through.

## Tests

- `tests/test_mcp_structured_error.py` — added `TestBuildMcpErrorResultContextEcho` (4 unit tests) covering: no params, params without context, params with context, sibling-not-nested placement; plus an end-to-end `test_context_echo_round_trips_through_register_tool`.
- `tests/test_a2a_structured_error.py` — added 3 executor-level tests covering: echo on raise, no-context omission, sibling-not-nested placement.
- All 29 MCP + 14 A2A structured-error tests pass; wider regression run on `tests/test_translate.py` + `tests/test_a2a_server.py` + `tests/test_server_helpers.py` (128 tests) green.

## Test plan

- [x] `pytest tests/test_mcp_structured_error.py tests/test_a2a_structured_error.py` — pass
- [x] `pytest tests/test_translate.py tests/test_a2a_server.py tests/test_server_helpers.py` — pass
- [x] `ruff check` — pass
- [x] pre-commit (black, ruff, mypy, bandit) — all hooks pass
- [ ] Storyboard runners pick up the fix on the 3 previously-failing assertions

### DIFF
--- a/src/adcp/server/a2a_server.py
+++ b/src/adcp/server/a2a_server.py
@@ -233,8 +233,13 @@ class ADCPAgentExecutor(AgentExecutor):
             # Binding, plus a human-readable text part. The JSON-RPC
             # channel is reserved for transport-level errors (auth
             # rejected, rate-limited pre-dispatch).
+            #
+            # ``params`` carries the raw request including any ``context``
+            # extension; threading it into the error builder echoes
+            # context back to the buyer on the error path, symmetric with
+            # the success path.
             logger.info("AdCP application error for skill %s: %s", skill_name, exc)
-            await self._send_adcp_error(event_queue, context, exc)
+            await self._send_adcp_error(event_queue, context, exc, params)
         except Exception:
             logger.exception("Error executing skill %s", skill_name)
             await self._send_error(event_queue, context, f"Skill execution failed: {skill_name}")
@@ -428,6 +433,7 @@ class ADCPAgentExecutor(AgentExecutor):
         event_queue: EventQueue,
         context: RequestContext,
         exc: Any,
+        params: dict[str, Any] | None = None,
     ) -> None:
         """Publish a failed task carrying an AdCP ``adcp_error`` payload.
 
@@ -442,9 +448,18 @@ class ADCPAgentExecutor(AgentExecutor):
         is shared with the MCP path via
         :func:`adcp.server.translate._extract_structured_fields`, so
         both transports project off the same source-of-truth shape.
+
+        When ``params`` is supplied and carries a wire ``context`` field,
+        that field is echoed alongside ``adcp_error`` in the DataPart —
+        symmetric with the success path's
+        :func:`adcp.server.helpers.inject_context` call. Without this
+        echo, error responses violate the AdCP context-passthrough
+        contract and buyers lose correlation IDs across the
+        raise-AdcpError boundary.
         """
         # Lazy import — ``translate.py`` pulls in heavier server deps
         # (mcp.types) which the A2A module doesn't otherwise need.
+        from adcp.server.helpers import inject_context
         from adcp.server.translate import _extract_structured_fields
 
         code, message, recovery, field, suggestion, details, _errors = _extract_structured_fields(
@@ -467,10 +482,14 @@ class ADCPAgentExecutor(AgentExecutor):
         if details:
             adcp_error["details"] = dict(details)
 
+        data: dict[str, Any] = {"adcp_error": adcp_error}
+        if params is not None:
+            inject_context(params, data)
+
         task = _make_task(
             context,
             state=pb.TaskState.TASK_STATE_FAILED,
-            data={"adcp_error": adcp_error},
+            data=data,
             message=message,
         )
         await event_queue.enqueue_event(task)
--- a/src/adcp/server/a2a_server.py
+++ b/src/adcp/server/a2a_server.py
@@ -226,18 +226,17 @@ class ADCPAgentExecutor(AgentExecutor):
             structured_error_types = (ADCPError, _DecisioningAdcpError)
         try:
             result = await self._dispatch_with_middleware(skill_name, params, tool_context)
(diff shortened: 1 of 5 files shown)


bokelley:
Updated based on expert review (code-reviewer + ad-tech-protocol-expert):

**Expanded scope**: code review caught that the A2A success path (`_send_result`) also skips `inject_context` — the original PR would have left a strange asymmetry (context echoed on A2A errors, dropped on A2A successes). Fixed in the same PR since it's the same bug class and a one-line addition; updated docstrings/tests accordingly.

**Other review fixes:**
- `build_mcp_error_result(exc, *, params=None)` is now keyword-only — locks the API shape against future positional drift.
- Added 64KB-cap drop test on the error path so the security note about buyer-controlled context size is enforced by a test, not just the docstring.
- Added A2A success-path echo + no-context-omission tests.

**Open question for reviewer**: ad-tech-protocol-expert flagged that `transport-errors.mdx` couldn't be fetched to confirm normative wire shape (sibling vs. nested). The PR matches the success-path convention (`context` at structuredContent / DataPart top level, not inside `adcp_error`), which is the safest reading. If upstream spec disagrees we'd want to align there before merging.

bokelley:
**On the open wire-shape question:** ran a second pass with `ad-tech-protocol-expert` specifically on sibling vs. nested placement. Verdict: **sibling is correct — proceed with confidence.**

Three reasons:

1. `inject_context()` is a top-level key injector by design (`response["context"] = ctx`). Calling it on `{"adcp_error": {...}}` is exactly the intended usage — there's no alternative injection point, and it guards against double-injection.
2. The success-path convention is `{"products": [...], "context": {...}}`. Placing `context` alongside `adcp_error` rather than inside it keeps the envelope shape symmetric across both outcomes: `context` is always a sibling of the payload key, never nested within it.
3. `adcp_error` is a typed schema object with defined fields (`code`, `message`, `recovery`, `field`, `suggestion`, `retry_after`, `details`). Injecting `context` inside it would mutate that schema to carry an opaque, variable-size blob — the schema-fidelity violation is the nesting case, not the sibling case.

One scoping note the expert raised (not a blocker): verify `params` in `_send_adcp_error` is the parsed wire-request dict that carries the `context` field, not a reduced/re-projected view. Checked the diff — `execute()` passes `params` directly through, so this is already correct.

The `transport-errors.mdx` fetch failure remains unresolved, but the symmetry argument is strong enough that this shouldn't block merge — if the spec later specifies nesting (which would be unusual given how context-passthrough is defined), it'd be a one-line fix.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01NAhicH2jcMyXpJe8E8gHsX

---

bokelley:
Two pre-merge follow-ups noted by @bokelley (filed as no-triage):

- **#562** — verify dispatcher non-AdcpError wrap path preserves context echo (any code path between handler and `serve.py:1798` that re-raises a non-`DecisioningAdcpError` skips the new echo).
- **#563** — `TaskHandoff` background-task failures route through `registry.fail` and aren't touched by this PR. Same fix shape needed there.

Shipping this PR now and tracking those two separately.

PR #564, opened 2026-05-04
fix(idempotency): support @wrap on arg-projected methods (update_media_buy)

### LINKED ISSUE #559: @IdempotencyStore.wrap incompatible with arg-projected methods (update_media_buy)
## What

`@IdempotencyStore.wrap` ([`src/adcp/server/idempotency/store.py:131-219`](https://github.com/adcontextprotocol/adcp-client-python/blob/main/src/adcp/server/idempotency/store.py#L131-L219)) wraps the handler with signature `(handler_self, params, context=None, *args, **kwargs)`. But the framework's arg-projector ([`src/adcp/decisioning/dispatch.py:~1080`](https://github.com/adcontextprotocol/adcp-client-python/blob/main/src/adcp/decisioning/dispatch.py)) splits some tools into per-field kwargs:

```python
# dispatch.py:1085 (arg_projector branch for update_media_buy)
result = await method(**arg_projector, ctx=ctx)
# arg_projector = {"media_buy_id": "...", "patch": {...}}
```

When `update_media_buy` is decorated with `@_IDEMPOTENCY.wrap`, the wrap body **never runs** — Python raises before:

```
TypeError: MockSellerPlatform.update_media_buy() missing 1 required positional argument: 'params'
```

Strictly a calling-convention incompatibility. Dispatch passes `media_buy_id=...`, `patch=...`, `ctx=...` as kwargs. None match the `params` positional, no default, the projected kwargs land in `**kwargs`, and Python errors out before `_wrapped`'s body executes. So this isn't a hash collision or cache-key issue — the wrap doesn't get a chance to compute anything.

## Repro

salesagent's `MockSellerPlatform.update_media_buy` decorated with `@_IDEMPOTENCY.wrap`. Method signature is `(self, media_buy_id, patch, ctx)` per the framework's arg-projector contract. Calling through MCP storyboard runner against `core/` raised the TypeError above on every `update_media_buy` request.

Reference: same shape as `examples/v3_reference_seller/src/platform.py:735` which intentionally does NOT wrap `update_media_buy` (presumably for the same reason — neither the example nor the docs spell this out as a known incompatibility).

## Workaround we shipped

Dropped `@_IDEMPOTENCY.wrap` from `update_media_buy` on both [`core/platforms/mock.py`](https://github.com/prebid/salesagent/blob/main/core/platforms/mock.py) and [`core/platforms/gam.py`](https://github.com/prebid/salesagent/blob/main/core/platforms/gam.py). **We've lost idempotency on update_media_buy** — buyer retries re-execute side effects. Acceptable for the spike; not for prod.

## What we'd want

**Cleanest fix on the wrap side:** make wrap signature-flexible so it works regardless of how the framework calls it. Detect the arg-projector calling pattern (kwargs that aren't `ctx`/`context`) and synthesize a params dict from those kwargs before hashing:

```python
async def _wrapped(handler_self, params=None, context=None, *args, **kwargs):
    if params is None and kwargs:
        # Arg-projector calling pattern — synthesize params from non-ctx kwargs
        ctx_kw = kwargs.pop("ctx", None) or kwargs.pop("context", None)
        params = kwargs
        if ctx_kw is not None:
            context = ctx_kw
    # ... existing logic, then dispatch back via the same calling pattern
    return await handler(handler_self, **params, ctx=context)  # for arg-projected
    # OR: return await handler(handler_self, params, context, *args, **kwargs)  # for normal
```

(Sketch only — needs tightening on which calling pattern was used so the inner `handler(...)` invocation matches.)

**Alternative on dispatch side:** detect a wrapped handler and route via `method(params=arg_projector_dict, ctx=ctx)` instead of unpacking. But dispatch is generic and shouldn't know about `@wrap`, so the wrap-side fix is cleaner.

## Filed by

salesagent kill-nginx/M2 spike — surfaced in [`core/SDK_FEEDBACK.md`](https://github.com/prebid/salesagent/blob/main/core/SDK_FEEDBACK.md) round 3 along with #558 (BearerTokenAuth × transport=both).

bokelley:
Closes #559

## Summary

`@IdempotencyStore.wrap` raised `TypeError: missing required positional argument: 'params'` when applied to methods the framework dispatches via the arg-projector path (e.g. `update_media_buy`). The arg-projector calls `method(media_buy_id=..., patch=..., ctx=ctx)` with no positional `params` — so the wrapper's `params: Any` parameter was never filled.

**Root cause:** the wrapper had no way to obtain the canonical wire-shape Pydantic model needed for RFC 8785 hashing and `idempotency_key` extraction, since the projected kwargs are a field-subset and cannot be reconstituted into a valid model.

**Fix approach:** dispatch now threads the original `params` object as `__adcp_params__=params` when `is_wrapped(method)` is True, giving the wrapper the full model without touching the normal call site. The wrapper pops the sentinel, detects the arg-projector convention (`params is None and len(field_kwargs) > 0`), and uses the canonical model for hashing while forwarding the original projected kwargs to the handler.

**Edge case fixed by pre-PR review:** direct adopter unit-test calls (e.g. `seller.update_media_buy(media_buy_id=..., ctx=ctx)` without dispatch or sentinel) previously would have fallen into `_prepare(None, ...)` and raised a confusing `TypeError` inside `_to_dict`. Now the wrapper falls through to the handler when `arg_projected=True` and the sentinel is absent, so wrapped methods are unit-testable without going through dispatch.

## Changes

- **`src/adcp/server/idempotency/store.py`** — `_MISSING` sentinel, updated `_wrapped` signature (`params: Any = None`), arg-projector detection, sentinel pop, sentinel-absent fall-through, `kwargs.clear()` for clarity
- **`src/adcp/decisioning/dispatch.py`** — `is_wrapped as _is_idem_wrapped` import; conditional `__adcp_params__=params` injection on both async and sync executor paths; inline comment on the sync path (forward-compat only, since `wrap` always returns an async def)
- **`tests/test_server_idempotency.py`** — 5 new tests in `TestArgProjectedCallingConvention`: no-TypeError, cache-hit dedup, payload conflict, no-key fall-through, and direct-call-without-sentinel fall-through

## Tested

- `pytest tests/test_server_idempotency.py` — 54 passed (5 new regression tests)
- `pytest tests/` (full suite) — 3,940 passed, 1 pre-existing failure (TLS network test unrelated to this change)
- `mypy src/adcp/` — 0 errors
- `ruff check src/` — 0 errors

## Pre-PR review

- **code-reviewer:** approved after blocker fix — blocker (direct-call without sentinel → confusing `TypeError` in `_prepare`) resolved by sentinel-absent fall-through; `kwargs = {}` → `kwargs.clear()` for clarity; sync executor dead path documented with inline comment. 1 nit surfaced below.
- **ad-tech-protocol-expert:** approved — non-breaking per AdCP #2315; RFC 8785 hash equivalence preserved (full wire-shape model hashed, not projected subset); per-principal scope isolation and context-echo contract unaffected; `idempotency_key` extraction from canonical model is sound.

**Nit (not blocking):** `__adcp_params__` is not a reserved kwarg at the Pydantic level — a request model with a field named `__adcp_params__` would have its field value shadowed at the call site by the sentinel. The double-underscore prefix is a reasonable deterrent; the sentinel name is documented in `store.py`'s module docstring as reserved in the arg-projector calling convention.

---

> issue). To unblock:
>
> - **Push fixup commits directly:** `gh pr checkout 560` →
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01HopGAcSiNk5SWbLpE9fd6b

---
_Generated by [Claude Code](https://claude.ai/code/session_01HopGAcSiNk5SWbLpE9fd6b)_

### DIFF
--- a/src/adcp/decisioning/dispatch.py
+++ b/src/adcp/decisioning/dispatch.py
@@ -1093,6 +1093,9 @@ async def _invoke_platform_method(
             ctx_snapshot = contextvars.copy_context()
             loop = asyncio.get_running_loop()
             if arg_projector is not None:
+                # _is_idem_wrapped(method) is always False here in practice
+                # (wrap() returns an async def), but the sentinel is wired in
+                # for forward-compatibility with a future sync-capable wrapper.
                 projected_kwargs = (
                     {**arg_projector, "ctx": ctx, "__adcp_params__": params}
                     if _is_idem_wrapped(method)
--- a/src/adcp/decisioning/dispatch.py
+++ b/src/adcp/decisioning/dispatch.py
@@ -64,6 +64,7 @@ from adcp.decisioning.types import (
     is_task_handoff,
     is_workflow_handoff,
 )
+from adcp.server.idempotency.store import is_wrapped as _is_idem_wrapped
 
 if TYPE_CHECKING:
     from collections.abc import Awaitable, Callable
@@ -1080,7 +1081,10 @@ async def _invoke_platform_method(
     try:
         if asyncio.iscoroutinefunction(method):
             if arg_projector is not None:
-                result = await method(**arg_projector, ctx=ctx)
+                if _is_idem_wrapped(method):
+                    result = await method(**arg_projector, ctx=ctx, __adcp_params__=params)
+                else:
+                    result = await method(**arg_projector, ctx=ctx)
             elif extra_kwargs:
                 result = await method(params, ctx, **extra_kwargs)
             else:
@@ -1089,7 +1093,11 @@ async def _invoke_platform_method(
             ctx_snapshot = contextvars.copy_context()
             loop = asyncio.get_running_loop()
             if arg_projector is not None:
-                projected_kwargs = {**arg_projector, "ctx": ctx}
+                projected_kwargs = (
+                    {**arg_projector, "ctx": ctx, "__adcp_params__": params}
+                    if _is_idem_wrapped(method)
+                    else {**arg_projector, "ctx": ctx}
+                )
                 result = await loop.run_in_executor(
                     executor,
                     functools.partial(ctx_snapshot.run, method, **projected_kwargs),
--- a/src/adcp/server/idempotency/store.py
+++ b/src/adcp/server/idempotency/store.py
@@ -176,9 +176,16 @@ class IdempotencyStore:
                 if kwargs:  # remaining kwargs are projected request fields
                     arg_projected = True
                     proj_kwargs = dict(kwargs)
-                    kwargs = {}
+                    kwargs.clear()
                     if ctx_val is not _MISSING:
                         context = ctx_val
+
+            # If the method was called directly (adopter unit-test, no dispatch
+            # sentinel), there is no params model to hash or extract a key from.
+            # Fall through so direct calls work without idempotency semantics
+            # rather than raising a confusing TypeError inside _prepare.
+            if arg_projected and canonical_params is _MISSING:
+                return await handler(handler_self, **proj_kwargs, ctx=context)
 
             # For hashing, prefer the canonical params object threaded from
             # dispatch (full wire-shape model); fall back to positional params
--- a/src/adcp/server/idempotency/store.py
+++ b/src/adcp/server/idempotency/store.py
@@ -59,6 +59,10 @@ logger = logging.getLogger(__name__)
 # only granted by IdempotencyStore.wrap itself.
 _WRAPPED_FUNCTIONS: weakref.WeakSet[Callable[..., Any]] = weakref.WeakSet()
 
+# Sentinel for kwarg presence detection — avoids truthiness bugs when the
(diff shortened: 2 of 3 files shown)


bokelley:
Superseded by #567 (merged today as 1abdae67), which closed #559 with the wrap-side resolution architecture rather than the dispatch-side sentinel approach this draft proposed. Per CLAUDE.md's parallel-race guidance for stale claude-triaged draft PRs: closing rather than rebasing, since the two architectures aren't compatible (#567 explicitly chose against threading dispatch→idempotency coupling). Issue #559 already in 4.4.0 changelog via #567.

bokelley:
Noted — supersedence by #567 (1abdae67) confirmed. Issue #559 is closed via the wrap-side resolution. No further action on this draft.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_01Rvm9f1JFJLjBfyvVrXVe1A_

---

PR #574, opened 2026-05-04
fix(auth): populate ctx.auth_principal from bearer ContextVar

### LINKED ISSUE #571: ctx.caller_identity is composite scope-key, not bare principal_id (docs misleading)
## What

The `RequestContext` docstring (adcp/decisioning/context.py:476-509) describes `auth_principal` as "who's calling" and `caller_identity` as "what's the cache scope key" / opaque. In practice for **bearer-token adopters**:

- `BearerTokenAuthMiddleware` populates `current_principal` ContextVar with the bare principal_id (whatever the adopter's `validate_token` returns on `Principal.caller_identity`)
- `auth_context_factory` reads that contextvar and sets `ToolContext.caller_identity` from it
- BUT — by the time downstream `_impl` code reads `ctx.caller_identity`, it's been **mutated to the composite scope key** `<store_module>.<store_qualname>:<account_id>` (e.g., `core.stores.accounts.SalesagentAccountStore:wonderstruck:default`)
- `ctx.auth_principal` stays `None` for bearer flows — that field only populates on signed-request flows where AuthInfo.principal is set

Net: a bearer-token adopter trying to read "who's calling?" from `ctx.caller_identity` gets the composite, not the bare id. salesagent's `_create_media_buy_impl` uses `identity.principal_id` for a Principal-table lookup, which 404'd on the composite key.

## Repro

```python
# core/platforms/_delegate.py
def _build_identity(ctx):
    # WRONG — returns composite scope-key in bearer flows
    principal_id = ctx.caller_identity  # → "core.stores.accounts.SalesagentAccountStore:wonderstruck:default"
    return ResolvedIdentity(principal_id=principal_id, ...)
```

The fix in salesagent was to import the framework's contextvar directly:

```python
from adcp.server.auth import current_principal
principal_id = current_principal.get()  # bare value, e.g. "wonderstruck_buyer"
```

That works but requires reaching into module-private state.

## What we'd want

Either:
1. Docs clarify that `caller_identity` is mutated to composite by downstream framework code, and document the correct read path for bearer adopters (`adcp.server.auth.current_principal.get()` or a typed helper)
2. Populate `ctx.auth_principal` from the bearer flow too — give bearer adopters a typed read for "who's calling" without reaching into ContextVars
3. Stop mutating `ctx.caller_identity` from its initial bare-id value; do composite-scoping inside the idempotency middleware via a separate field

(2) is probably the cleanest — `auth_principal` already exists as the "typed convenience field"; populating it on bearer flows matches its docstring intent.

## Filed by

salesagent live GAM CRUD probe — surfaced when `_create_media_buy_impl` Principal lookup failed because identity.principal_id was the composite. Tracked as salesagent #42.

bokelley:
## Summary

Closes #571.

For bearer-token adopters using `BearerTokenAuthMiddleware`, `RequestContext.auth_principal` was always `None` because the middleware never constructs an `AuthInfo` — the principal lives in the `adcp.server.auth.current_principal` ContextVar that `auth_context_factory` reads into `ToolContext.caller_identity`. Adopters trying to read "who's calling?" inside a decisioning handler either had to reach into the framework-private ContextVar themselves, or misread `ctx.caller_identity` (which the framework dispatch helper mutates downstream into a composite cache scope key for idempotency, e.g. `core.stores.accounts.SalesagentAccountStore:wonderstruck:default`).

Three coordinated parts:

- **A. Code fix.** `_build_request_context` in `src/adcp/decisioning/dispatch.py` now falls back to `current_principal.get()` when `auth_info` is absent or its `principal` is `None`. Local import inside the function avoids the server→decisioning import cycle. Outside a bearer flow `current_principal.get()` returns `None` — the desired no-op for unauthenticated dev fixtures and signed-request flows that already pass `AuthInfo`.
- **B. Docstrings.** `RequestContext` docstring in `src/adcp/decisioning/context.py` clarifies that `auth_principal` is sourced from `AuthInfo.principal` on signed-request flows AND the bearer ContextVar on bearer flows, and that `caller_identity` is mutated downstream into a composite cache scope key — adopters should not read it for identity decisions. `_build_request_context` param doc updated to match.
- **C. Examples + docs.** `examples/mcp_with_auth_middleware.py` carries an inline comment pointing at `ctx.auth_principal` as the right read for "who's calling?". `docs/handler-authoring.md` gains a "Reading 'who's calling?' — `auth_principal` vs `caller_identity`" subsection that disambiguates the two fields and their layers.

## Behavior shift (call out)

Bearer-flow callers that previously saw `ctx.auth_principal == None` will now see the bearer principal (e.g. `principal-acme-ops`). Adopter code that gated on `if ctx.auth_principal is None: ...` for bearer flows will silently change behavior. This is intended — the previous `None` was the bug filed against the SDK by the known bearer adopter (salesagent) — but worth flagging for any other bearer adopter on `is None` checks.

`auth_info` is unchanged for bearer flows: still `None`. Only `auth_principal` is populated from the ContextVar fallback. The deeper structural gap (`BearerTokenAuthMiddleware` not constructing an `AuthInfo`) is a separate follow-up flagged by the security-reviewer in triage.

## Test plan

- [x] `pytest tests/test_decisioning_dispatch.py tests/test_decisioning_context_state_resolve.py tests/test_testing_decisioning.py tests/test_auth_middleware.py tests/test_serve_auth_both.py -v` — 152 passed
- [x] `pytest tests/ -v -k "auth or dispatch or context"` — 566 passed, 12 skipped, 1 xfailed
- [x] `ruff check src/` — clean
- [x] `mypy src/adcp/` — clean
- [x] New regression test `test_build_request_context_falls_back_to_bearer_context_var` exercises the new fallback path explicitly
- [x] New test `test_build_request_context_auth_info_takes_precedence_over_bearer_var` pins the precedence rule (signed-request `AuthInfo` wins when both sources are populated)
- [ ] Manual smoke from a bearer adopter confirming `ctx.auth_principal` now reads the principal in their handler

### DIFF
--- a/docs/handler-authoring.md
+++ b/docs/handler-authoring.md
@@ -295,6 +295,30 @@ and leak the presence of an auth path to attackers).
 Full worked example: `examples/mcp_with_auth_middleware.py`. Integration
 test proving the composition: `tests/test_mcp_middleware_composition.py`.
 
+#### Reading "who's calling?" — `auth_principal` vs `caller_identity`
+
+The two identity-shaped fields adopters see have distinct purposes and
+the wrong read silently misroutes authorization decisions.
+
+- **`ctx.auth_principal`** — the verified caller's identity label.
+  Read this for per-principal ACLs ("can principal X mutate this
+  buy?"). Sourced from `AuthInfo.principal` on signed-request flows
+  and from the `current_principal` ContextVar on bearer flows.
+  `None` for unauthenticated dev fixtures.
+- **`ctx.caller_identity`** — a cache scope key, not a principal.
+  At the transport layer (`ToolContext.caller_identity`) it's the
+  bare principal; the framework dispatch helper mutates it into a
+  composite key (`<store_module>.<store_qualname>:<account_id>`)
+  before the handler sees a decisioning `RequestContext`. The
+  idempotency middleware reads the composite form. Treat as opaque
+  inside dispatch handlers — log or forward, but don't parse,
+  compare, or rewrite.
+
+Skill-middleware and `context_factory` code (which run at the
+transport layer, before dispatch hydration) still read
+`context.caller_identity` for legitimate cache / rate-limit keying;
+the composite mutation happens later, in `_build_request_context`.
+
 #### Pattern 2a — custom middleware (when the shipped one doesn't fit)
 
 Subclass `BearerTokenAuthMiddleware` to tighten the discovery bypass,
--- a/examples/mcp_with_auth_middleware.py
+++ b/examples/mcp_with_auth_middleware.py
@@ -63,6 +63,15 @@ class MultiTenantSalesAgent(ADCPHandler):
         return capabilities_response(["media_buy"])
 
     async def get_products(self, params: Any, context: ToolContext | None = None) -> dict[str, Any]:
+        # On the dispatch path the handler receives a decisioning
+        # ``RequestContext``; read ``ctx.auth_principal`` (not
+        # ``ctx.caller_identity``) for "who's calling?". The
+        # framework mutates ``caller_identity`` downstream into a
+        # composite cache scope key for idempotency — it is not a
+        # principal label by the time a handler sees it. On bearer
+        # flows like this one ``auth_principal`` is sourced from the
+        # :data:`adcp.server.auth.current_principal` ContextVar that
+        # the middleware populates.
         tenant = context.tenant_id if context is not None else None
         return products_response(_products_for_tenant(tenant))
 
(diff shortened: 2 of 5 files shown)


bokelley:
Reworded the local-import justification comment in `src/adcp/decisioning/dispatch.py` per review feedback. The previous comment claimed the local import sidesteps a `server→decisioning` import cycle, but `adcp.server.auth` does not import from `adcp.decisioning` (verified its imports), so there is no real cycle. The local-import pattern is still correct — it just avoids forcing a top-level dep on `adcp.server.auth` from `dispatch.py` — and the comment now says that instead.

Comment-only change; no behavior, no code movement.

bokelley:
Confirmed — `073beec` is visible on the branch. The updated comment ("Local import keeps the layering local — read the bearer ContextVar without forcing a top-level dep on `adcp.server.auth`") is accurate and avoids the misleading cycle claim. No further action needed here.

---

bokelley:
Acknowledged — thanks for confirming. No further action from me; standing by for review.

---

PR #591, opened 2026-05-05
fix(server): propagate auth ContextVars through A2A dispatch boundary

### LINKED ISSUE #590: A2A: `current_principal` contextvar is `None` inside platform handlers (`current_tenant` works) — same token works on MCP
**Affected versions:** `adcp==4.4.0`, `4.4.1`, `4.4.2`. Reproduces against both `a2a-sdk==1.0.1` and `1.0.2` — not specific to either, this is about contextvar propagation through the A2A request-handler dispatch boundary.

## Symptom

Same bearer token, same handler, same `DecisioningPlatform`. Two transports, two outcomes:

| Transport | Call | Result |
|---|---|---|
| MCP `tools/call` | `get_products` | `{"products": []}` ✅ |
| A2A `message/send` | `get_products` | `AdCPAuthenticationError("Authentication required by tenant policy")` ❌ |

The platform method's view: `current_principal.get()` returns `None` only on the A2A path. `current_tenant.get()` works fine on both — only the principal contextvar drops.

## Where it breaks

`adcp.server.auth.BearerTokenAuthMiddleware.dispatch` sets the contextvar before calling downstream:

```python
# adcp/server/auth.py:308 (4.4.x)
principal_token = current_principal.set(principal.caller_identity)
tenant_token = current_tenant.set(principal.tenant_id)
metadata_token = current_principal_metadata.set(...)
return await call_next(request)
```

On MCP this propagates fine — the platform method sees `current_principal.get() == "<caller_identity>"`. On A2A, by the time `platform.get_products(req, ctx)` runs, `current_principal.get() == None` while `current_tenant.get()` is still populated. Different propagation behavior on the same Starlette binary, same middleware, same call.

Likely culprit: the A2A request handler in `a2a-sdk` runs the actual skill dispatch on a different async task / cancel scope than the MCP transport, and only one of the two contextvars survives the boundary. Tenant survives because it's also set deeper in the stack (`SubdomainTenantMiddleware`); principal is only ever set by `BearerTokenAuthMiddleware`, so when that contextvar is lost there's no fallback.

## Minimal repro (`adcp` + `a2a-sdk` only)

```python
# server.py
from adcp.decisioning import DecisioningPlatform, serve
from adcp.decisioning.capabilities import DecisioningCapabilities, MediaBuy, SupportedProtocol
from adcp.server import BearerTokenAuth, Principal
from adcp.server.auth import current_principal, current_tenant


class DiagPlatform(DecisioningPlatform):
    capabilities = DecisioningCapabilities(
        media_buy=MediaBuy(supported_pricing_models=["cpm"]),
        supported_protocols=[SupportedProtocol.media_buy],
    )

    async def get_products(self, req, ctx):
        return {
            "products": [],
            "_diag": {
                "current_principal": current_principal.get(),
                "current_tenant": current_tenant.get(),
            },
        }


def validate(token: str) -> Principal | None:
    if token == "good-token":
        return Principal(caller_identity="alice", tenant_id="t1")
    return None


serve(
    DiagPlatform(),
    transport="both",                       # MCP at /mcp, A2A at /
    auth=BearerTokenAuth(
        validate_token=validate,
        header_name="x-adcp-auth",
        bearer_prefix_required=False,
    ),
)
```

```bash
# A2A
curl -s -X POST http://localhost:3001/ \
  -H "x-adcp-auth: good-token" -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"message/send","id":"a","params":{"message":{"role":"user","messageId":"m","kind":"message","parts":[{"kind":"data","data":{"skill":"get_products","parameters":{}}}]}}}'
# Expected: _diag.current_principal == "alice"
# Actual:   _diag.current_principal == null

# MCP (same token, same skill)
curl -s -X POST http://localhost:3001/mcp/ \
  -H "x-adcp-auth: good-token" -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"tools/call","id":"m","params":{"name":"get_products","arguments":{}}}'
# Expected and Actual: _diag.current_principal == "alice"
```

## Live observation

From a `salesagent` deployment running `adcp 4.4.2 + a2a-sdk 1.0.1 + protobuf 6.33.6`:

```
adcp-server-1  | 2026-05-05 19:46:32 - src.core.tools.products - INFO - [GET_PRODUCTS] Tenant context: default
adcp-server-1  | 2026-05-05 19:46:32 - adcp.decisioning.dispatch - ERROR - Unhandled exception in platform.get_products
adcp-server-1  |   File ".../core/platforms/_delegate.py", line 87, in _build_identity
adcp-server-1  |     principal_id = current_principal.get() or getattr(ctx, "auth_principal", None)
adcp-server-1  |   File ".../src/core/tools/products.py", line 201, in _get_products_impl
adcp-server-1  |     raise AdCPAuthenticationError("Authentication required by tenant policy")
```

`Tenant context: default` → `current_tenant` survives.
`AuthenticationError` from `principal_id is None` → `current_principal` does not.

Falling back to `getattr(ctx, "auth_principal", None)` doesn't help — that attribute also isn't populated on A2A requests.

## Why it matters

Every adopter with a tenant policy of `require_auth` (the salesagent default) gets locked out of A2A entirely while MCP works. Sellers that tighten policy to require principal-bound calls have no working A2A surface today. This is why we wired our integration to read `current_principal.get()` directly — the docstring on `BearerTokenAuthMiddleware` advertises that as the primary read path:

> On success, populates `current_principal`, `current_tenant`, and `current_principal_metadata` for the duration of the downstream call.

That contract holds on MCP and breaks on A2A.

## Asks

1. Confirm whether `current_principal` is supposed to propagate through the A2A request-handler boundary, or whether adopters need to read identity from `ctx.auth_principal` / `ctx.account.metadata` instead. (We already tried `ctx.auth_principal` — not populated on A2A.)
2. **If the contextvar boundary is the bug:** keep `current_principal` set across the dispatch boundary the same way `current_tenant` is — most likely by re-binding the contextvar inside the A2A skill dispatcher (or before the task it spawns) using the value the middleware computed.
3. **If the recommended path is `ctx.auth_principal`:** document that `current_principal.get()` is unreliable on A2A and add a stable, populated attribute on `RequestContext` that adopters can read in `platform.<method>` regardless of transport. Today neither side of the API is reliable on A2A.

Tagging the salesagent deployment we hit this on: https://github.com/prebid/salesagent (current branch: `bokelley/wrapper-webhook-fix`, PR #17).

bokelley:
Closes #590

## Summary

`A2ABearerAuthMiddleware` validated bearer tokens on the A2A transport but only stored the authenticated `Principal` in `scope["user"]` and `scope["auth"]`. It never set the module-level ContextVars (`current_principal`, `current_tenant`, `current_principal_metadata`) that `BearerTokenAuthMiddleware` sets on the MCP path. As a result, `current_principal.get()` returned `None` inside every A2A handler call even after successful auth — breaking the documented contract ("On success, populates `current_principal`, `current_tenant`, and `current_principal_metadata` for the duration of the downstream call") for the A2A leg.

Root effect: every adopter whose platform method reads `current_principal.get()` directly, or whose `require_auth` tenant policy depends on `auth_context_factory` populating `AuthInfo`, was locked out of A2A while MCP worked fine.

The fix mirrors the existing `BearerTokenAuthMiddleware` try/finally pattern:
- On auth success: set all three ContextVars from the validated `Principal`, reset unconditionally in `finally`.
- On OPTIONS/discovery pass-throughs: set all three to `None` (clearing any stale value from an enclosing task context), reset in `finally`. Matches the MCP middleware's discovery branch.
- On non-HTTP scopes (lifespan, websocket): pass through unchanged — skill handlers don't run on these paths.

## What was tested

- `pytest tests/test_serve_auth_both.py` — **30 passed** (including 2 new regression tests)
- `pytest tests/test_auth_middleware.py tests/test_a2a_server.py tests/test_decisioning_dispatch.py` — **123 passed**
- Full non-integration suite: **4120 passed, 0 failed** (1 pre-existing network flake in `test_ip_pinned_transport.py::test_real_tls_handshake_still_validates_hostname` excluded — hits `example.com`, unrelated)
- `ruff check src/adcp/server/auth.py` — clean
- `mypy src/adcp/server/auth.py --ignore-missing-imports` — no new errors (pre-existing `BaseHTTPMiddleware` stub issue at line 197 unchanged)

**Pre-PR review:**
- code-reviewer: approved — no blockers; nits: `current_principal_metadata` not asserted in unit test (coverage gap, not a correctness risk — set/reset logic is structurally identical for all three tokens); OPTIONS/discovery reset-in-finally not explicitly tested (would require inner app to raise, and ContextVar resets to `None` which is also the default)
- security-reviewer: approved — 401 path (inside try before any token set) confirmed safe: finally is a no-op when all tokens are `None`; clearing to `None` on OPTIONS/discovery confirmed correct and matching the MCP middleware's discovery branch; no new attack surface

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01ETystgT4HH4ZHaAziszjBF

---
_Generated by [Claude Code](https://claude.ai/code/session_01ETystgT4HH4ZHaAziszjBF)_

### DIFF
--- a/src/adcp/server/auth.py
+++ b/src/adcp/server/auth.py
@@ -637,41 +637,60 @@ class A2ABearerAuthMiddleware:
 
     async def __call__(self, scope: Any, receive: Any, send: Any) -> None:
         # Lifespan + websocket pass through unchanged. Auth applies to
-        # HTTP requests only.
+        # HTTP requests only. No contextvar changes — these scopes are
+        # not dispatched to skill handlers that read auth vars.
         if scope.get("type") != "http":
             await self._app(scope, receive, send)
             return
 
-        # CORS preflight is part of the public surface — browser-origin
-        # clients send ``OPTIONS`` before any auth'd POST. Returning 401
-        # here breaks the preflight and the buyer never gets a chance to
-        # retry with a token. Pass through; let the inner app's CORS
-        # handler (or operator-supplied ``asgi_middleware``) respond.
-        if scope.get("method") == "OPTIONS":
+        principal_token = None
+        tenant_token = None
+        metadata_token = None
+        try:
+            # CORS preflight and A2A discovery are part of the public surface.
+            # Set contextvars to None to prevent stale values from an enclosing
+            # task context from leaking into downstream code on these paths —
+            # mirrors BearerTokenAuthMiddleware's discovery branch.
+            if scope.get("method") == "OPTIONS" or scope.get("path", "") in _A2A_DISCOVERY_PATHS:
+                principal_token = current_principal.set(None)
+                tenant_token = current_tenant.set(None)
+                metadata_token = current_principal_metadata.set(None)
+                await self._app(scope, receive, send)
+                return
+
+            principal = self._authenticate_scope(scope)
+            if principal is None:
+                await self._send_unauthenticated(send)
+                return
+
+            # Stash both the duck-typed user (for DefaultServerCallContextBuilder)
+            # and the raw Principal (for downstream code reading scope['auth']).
+            # Mutating the scope dict before delegating propagates state to
+            # nested apps without copying.
+            scope["user"] = _A2AAuthenticatedUser(
+                display_name=principal.caller_identity,
+                tenant_id=principal.tenant_id,
+                principal_metadata=dict(principal.metadata) if principal.metadata else None,
+            )
+            scope["auth"] = principal
+            # Populate the same module-level ContextVars that BearerTokenAuthMiddleware
+            # sets on the MCP path. auth_context_factory and adopter code that reads
+            # current_principal.get() directly see the authenticated identity on A2A
+            # exactly as they do on MCP. Reset unconditionally in finally so a later
+            # task sharing this context can't read a stale principal.
+            principal_token = current_principal.set(principal.caller_identity)
+            tenant_token = current_tenant.set(principal.tenant_id)
+            metadata_token = current_principal_metadata.set(
+                dict(principal.metadata) if principal.metadata else None
+            )
             await self._app(scope, receive, send)
-            return
-
-        path = scope.get("path", "")
-        if path in _A2A_DISCOVERY_PATHS:
-            await self._app(scope, receive, send)
-            return
-
-        principal = self._authenticate_scope(scope)
-        if principal is None:
-            await self._send_unauthenticated(send)
-            return
-
-        # Stash both the duck-typed user (for DefaultServerCallContextBuilder)
-        # and the raw Principal (for downstream code reading scope['auth']).
-        # Mutating the scope dict before delegating propagates state to
-        # nested apps without copying.
-        scope["user"] = _A2AAuthenticatedUser(
-            display_name=principal.caller_identity,
-            tenant_id=principal.tenant_id,
-            principal_metadata=dict(principal.metadata) if principal.metadata else None,
-        )
-        scope["auth"] = principal
-        await self._app(scope, receive, send)
+        finally:
+            if principal_token is not None:
+                current_principal.reset(principal_token)
+            if tenant_token is not None:
+                current_tenant.reset(tenant_token)
+            if metadata_token is not None:
+                current_principal_metadata.reset(metadata_token)
 
     def _authenticate_scope(self, scope: Any) -> Principal | None:
         """Read + validate the bearer header off raw ASGI scope.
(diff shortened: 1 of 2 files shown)


bokelley:
Closing as superseded by #592, which shipped the same A2A auth middleware ContextVar fix to main (commit 54309427). The A2ABearerAuthMiddleware on current main already populates `current_principal`, `current_tenant`, and `current_principal_metadata` from the validated Principal.

PR #609, opened 2026-05-09
fix(decisioning): wire sync_accounts/list_accounts dispatch to AccountStore Protocols

bokelley:
## Problem

`PlatformHandler.sync_accounts` and `PlatformHandler.list_accounts` are inherited as `_not_supported` stubs from `ADCPHandler`. Every adopter that implements `AccountStoreUpsert` / `AccountStoreList` on their platform's `accounts` store is invisible on the wire — both tools return `OPERATION_NOT_SUPPORTED` regardless of what the store declares.

Compared to e.g. `provide_performance_feedback` in the same handler, which properly threads through `_invoke_platform_method` to `platform.provide_performance_feedback(params, ctx)`. The account dispatchers were never given the equivalent treatment.

Tracked at [adcontextprotocol/adcp-client#1631](https://github.com/adcontextprotocol/adcp-client/issues/1631).

## Fix

Wire real shims that route through the documented design — `AccountStore.upsert` / `.list`, NOT `platform.<method>`, since `LazyPlatformRouter._ACCOUNT_STORE_METHODS` already excludes account ops from per-tenant delegation:

```python
async def sync_accounts(self, params, context=None):
    upsert = getattr(self._platform.accounts, "upsert", None)
    if upsert is None:
        return self._not_supported("sync_accounts")
    tool_ctx = context or ToolContext()
    account = await self._resolve_account(None, tool_ctx)
    ctx = self._build_ctx(tool_ctx, account)
    resolve_ctx = self._make_resolve_context(tool_ctx, "sync_accounts")
    result = _call_with_optional_ctx(upsert, params, ctx=resolve_ctx)
    if inspect.isawaitable(result):
        result = await result
    return cast("SyncAccountsResponse", result)
```

The `ResolveContext` carries the caller's `AuthInfo` and resolved `BuyerAgent` — same threading `AccountStore.resolve` already uses, so adopter impls implementing principal-keyed gates (e.g. spec `BILLING_NOT_PERMITTED_FOR_AGENT`) read the principal off the canonical context.

Both `upsert` and `list` are OPTIONAL on the AccountStore. The shims surface `OPERATION_NOT_SUPPORTED` (via `_not_supported`) when the store doesn't expose them — distinct from `AttributeError`, which is what an unguarded `getattr().()` would produce.

## Wire advertisement

Adds `_ACCOUNT_ADVERTISED_TOOLS = {"sync_accounts", "list_accounts"}` and unions it into every `sales-*` entry of `SPECIALISM_TO_ADVERTISED_TOOLS` (every sales-* claim implies an account roster — per spec, buyers must be able to reach it). The override-detection filter still drops the tools when the platform's store doesn't expose the corresponding Protocol method, so adopters who don't yet implement them aren't suddenly over-advertising.

## Tests

- `test_advertised_tools_covers_every_specialism_wire_tool` — extended with the two new tools.
- New positive: `sync_accounts` / `list_accounts` route through to `platform.accounts.upsert` / `.list` with `ResolveContext.tool_name` set correctly.
- New negative: stores that don't implement the optional Protocols surface `OPERATION_NOT_SUPPORTED` via `NotImplementedResponse` — distinct from `AttributeError` from an unguarded `getattr` chain.

877 tests pass locally (`pytest tests/ -k "decisioning or account or handler or platform"`).

## Storyboard impact

Downstream sellers running the AdCP storyboard `media_buy_seller/refine_products/setup` against an agent that implements the optional store Protocols will see the scenario graduate from `skipped (missing_tool)` to graded the moment they pull this in. Verified end-to-end against [bokelley/salesagent@4e62024c](https://github.com/bokelley/salesagent/commit/4e62024c) — `MCP 31→32 passed, 28→27 skipped`.

## Side note

Tracked salesagent monkey-patch that this PR replaces: https://github.com/bokelley/salesagent/blob/main/core/platforms/account_polyfill.py — happy to drop the polyfill once this lands.

### DIFF
--- a/src/adcp/decisioning/handler.py
+++ b/src/adcp/decisioning/handler.py
@@ -81,6 +81,7 @@ logger = logging.getLogger(__name__)
 # exception, and the dispatcher falls back to the dict path — which
 # crashes inside the shim with ``'dict' object has no attribute
 # 'account'`` (Emma sales-direct backend test, verdict 2/10).
+from adcp.decisioning.accounts import ResolveContext, _call_with_optional_ctx
 from adcp.types import (
     AccountReference,
     AcquireRightsRequest,
@@ -131,6 +132,8 @@ from adcp.types import (
     GetRightsSuccessResponse,
     GetSignalsRequest,
     GetSignalsResponse,
+    ListAccountsRequest,
+    ListAccountsResponse,
     ListCollectionListsRequest,
     ListCollectionListsResponse,
     ListContentStandardsRequest,
@@ -147,6 +150,8 @@ from adcp.types import (
     ProvidePerformanceFeedbackResponse,
     ReportPlanOutcomeRequest,
     ReportPlanOutcomeResponse,
+    SyncAccountsRequest,
+    SyncAccountsResponse,
     SyncAudiencesRequest,
     SyncAudiencesSuccessResponse,
     SyncCreativesRequest,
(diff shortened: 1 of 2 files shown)


bokelley:
Superseded by #610 — same core wiring, but strictly better: per-instance filter that drops the tool when the store doesn't expose the optional Protocol (with logger.info breadcrumb), wire-envelope auto-wrapping with credential scrubbing, filter-dict projection for list, plus authoring docs and credential-smuggling test coverage. Closing in favor.

bokelley:
Acknowledged — noting the closure in favor of #610. No action needed here.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_0146e4vWbsnkKQ7WqUvLw3ue

---

PR #613, opened 2026-05-10
fix(server): register /.well-known/agent.json alias route in create_a2a_server

### LINKED ISSUE #612: A2A: /.well-known/agent.json (0.3 alias) returns 404 — route not registered, breaking SDK auto-detection
## Summary

The Python SDK's A2A server registers a Starlette route for the canonical `/.well-known/agent-card.json` (via `a2a-sdk`'s `create_agent_card_routes`) but **does not register a route for the 0.3-compat alias `/.well-known/agent.json`**. Any healthy `adcp-client-python` agent therefore returns 404 (or 503 at edge proxies that don't return application 404 from no-route conditions, e.g. Fly.io) on the alias.

This breaks `@adcp/sdk` (JS) CLI auto-detection of A2A transport — the JS SDK probes both well-known paths and uses the alias as a positive A2A signal. Reported by users running `npx @adcp/sdk@6.15.0 storyboard run <agent>` against a Python-SDK-built seller — A2A transport not detected, despite the agent being fully functional on the canonical path.

## Repro

Any agent built on `adcp-client-python` with `serve(transport=\"a2a\", ...)` or `serve(transport=\"both\", ...)`:

\`\`\`bash
curl -i 'https://<agent>/.well-known/agent-card.json'    # 200 OK (canonical, works)
curl -i 'https://<agent>/.well-known/agent.json'         # 404 (or 503 at edge — route never registered)
\`\`\`

Live example: `https://wonderstruck.sales-agent.scope3.com/.well-known/agent.json` → HTTP 503 from Fly's edge.

## Root cause

`adcp/server/a2a_server.py:833`:

\`\`\`python
routes = list(create_agent_card_routes(agent_card=agent_card)) + list(
    create_jsonrpc_routes(**jsonrpc_kwargs)
)
\`\`\`

`create_agent_card_routes` (from `a2a-sdk`) registers exactly **one** route at `AGENT_CARD_WELL_KNOWN_PATH = '/.well-known/agent-card.json'`. The 0.3 alias path is mentioned only defensively elsewhere — `adcp/server/auth.py:603-604` exempts both paths from auth (\"Legacy 0.3 alias retained by enable_v0_3_compat=True\"), but the auth list doesn't *create* routes, it only *exempts* them.

`enable_v0_3_compat=True` (line 829) is passed to `create_jsonrpc_routes` (which dual-serves 0.3/1.0 RPC formats on `/`) but is **not** plumbed through to the agent-card route construction — the flag's promise of \"0.3-compat retention\" is incomplete for discovery.

## Why it matters

The 0.3 alias is documented in your code as \"retained\" by `enable_v0_3_compat=True`. Auth was wired for it. Buyer SDKs still probe it. But the route was never registered. This makes `enable_v0_3_compat` partially load-bearing in a way that surprises adopters: their 0.3 RPC traffic works, their 0.3 discovery silently breaks. SDK auto-detection logic that uses the alias as a transport hint (as `@adcp/sdk`'s does) treats Python-SDK agents as if they don't speak A2A.

## Suggested fix

In `adcp/server/a2a_server.py:833`, when `enable_v0_3_compat=True` is in effect, append a redirect or duplicate route at `/.well-known/agent.json`:

\`\`\`python
from starlette.responses import RedirectResponse
from starlette.routing import Route

# ... after create_agent_card_routes(...) ...

if enable_v0_3_compat:
    async def _redirect_v0_3_alias(request):
        return RedirectResponse('/.well-known/agent-card.json', status_code=308)

    routes.append(Route(
        '/.well-known/agent.json',
        endpoint=_redirect_v0_3_alias,
        methods=['GET'],
    ))
\`\`\`

308 (vs 301/302) preserves request method and is the modern correct status for permanent redirects to a different URL where the original is deprecated. Buyer SDKs that follow redirects (which both `@adcp/sdk` and most HTTP clients do by default) get the canonical agent-card content transparently.

Alternative: register a duplicate route handler that returns the same body. Slightly more bytes per request, no redirect round-trip. Either is a valid resolution.

## Workaround (already shipped on our agent)

We shipped an ASGI redirect middleware on our agent as the immediate workaround: https://github.com/bokelley/salesagent/pull/269 (closes bokelley/salesagent#267). After this lands upstream, we'd remove our middleware as cleanup.

## Tests

A regression test in `tests/server/test_a2a_server.py` (or wherever the agent-card route tests live) that asserts:

1. GET `/.well-known/agent.json` returns 308 (or 200 if you choose the duplicate-handler path).
2. Following the redirect resolves to a 200 with valid agent-card JSON matching the canonical path.

## Environment

- `adcp-client-python`: pinned version per salesagent's `uv.lock` as of 2026-05-09
- AdCP spec: 3.0.8
- a2a-sdk: 1.x

bokelley:
Closes #612

## Summary

`create_agent_card_routes()` (a2a-sdk) registers exactly one route: `/.well-known/agent-card.json`. The 0.3 discovery alias `/.well-known/agent.json` was already present in `_A2A_DISCOVERY_PATHS` (auth middleware exempted it) but was never registered as a Starlette route — returning 404 or 503 at edge proxies. This broke `@adcp/sdk` CLI auto-detection of A2A transport, which probes the alias as a positive A2A signal.

**Fix:** Call `create_agent_card_routes(agent_card=agent_card, card_url="/.well-known/agent.json")` a second time in `create_a2a_server()`. Both paths serve identical agent-card JSON using independent handler closures — no redirect round-trip, no shared mutable state.

**Also fixed:**
- Three tests that masked the bug by accepting 404 as valid (`test_unified_mcp_a2a.py`) or asserting only the canonical path (`test_a2a_server.py`)
- New inverse coverage test (`test_all_discovery_paths_are_registered_routes`) asserting every path in `_A2A_DISCOVERY_PATHS` has a live Starlette route — this is the structural check that would have caught issue #612 at the point `/.well-known/agent.json` was added to the frozenset
- Misleading comment in `auth.py` ("retained by enable_v0_3_compat=True") corrected to note the route is registered explicitly in `create_a2a_server`
- `A2ABearerAuthMiddleware` docstring updated to mention both exempted paths
- `docs/handler-authoring.md` corrected to name the canonical path (`/.well-known/agent-card.json`) with a parenthetical noting the 0.3 alias

**Nits (not fixed in this PR):**
- `test_a2a_server.py:349` uses `startswith("/.well-known/agent-card")` — could use `AGENT_CARD_WELL_KNOWN_PATH` constant for exactness

## What was tested

```
ruff check src/           # All checks passed
pytest tests/test_a2a_server.py tests/test_unified_mcp_a2a.py tests/test_serve_auth_both.py -v
# 80 passed, 0 failed, 43 warnings
```

## Pre-PR review

- **code-reviewer:** approved — no blockers. Core fix is correct (`create_agent_card_routes` creates fresh independent closures per call; no shared state). Two issues addressed (auth docstring, test limitation comment). Noted that `test_unified_mcp_a2a.py` tightening from `in (200, 404)` → `== 200` is the right move.
- **ad-tech-protocol-expert:** approved — no blockers. `/.well-known/agent.json` is confirmed as the intended 0.3 alias (already in `_A2A_DISCOVERY_PATHS` with intentional auth exemption). No canonicalization concern (neither well-known path appears in the serialized card body). Absence of `Link: rel=canonical` on alias response is consistent with all other A2A implementations and not required by the spec.

---

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01JHq6DLPhFf2kQQpUkPtfZe

---
_Generated by [Claude Code](https://claude.ai/code/session_01JHq6DLPhFf2kQQpUkPtfZe)_

### DIFF
--- a/docs/handler-authoring.md
+++ b/docs/handler-authoring.md
@@ -929,8 +929,9 @@ mock-B).
 ## A2A transport
 
 `serve(MyAgent(), transport="a2a")` wires the same handler through the
-A2A protocol with auto-generated agent card (`/.well-known/agent.json`)
-derived from the `ADCPHandler` methods your class overrides.
+A2A protocol with auto-generated agent card (`/.well-known/agent-card.json`,
+with the 0.3 alias `/.well-known/agent.json` also served for backwards
+compatibility) derived from the `ADCPHandler` methods your class overrides.
 
 ### Durable task storage
 
--- a/src/adcp/server/a2a_server.py
+++ b/src/adcp/server/a2a_server.py
@@ -919,7 +919,7 @@ def create_a2a_server(
         jsonrpc_kwargs["context_builder"] = context_builder
     routes = (
         list(create_agent_card_routes(agent_card=agent_card))
-        # 0.3 alias: buyer SDKs (e.g. @adcp/sdk) probe /.well-known/agent.json
+        # 0.3 alias: A2A 0.3 buyer SDKs probe /.well-known/agent.json
         # as a positive A2A signal. Same handler, no redirect round-trip.
         + list(
             create_agent_card_routes(
--- a/src/adcp/server/a2a_server.py
+++ b/src/adcp/server/a2a_server.py
@@ -917,8 +917,16 @@ def create_a2a_server(
     }
     if context_builder is not None:
         jsonrpc_kwargs["context_builder"] = context_builder
-    routes = list(create_agent_card_routes(agent_card=agent_card)) + list(
-        create_jsonrpc_routes(**jsonrpc_kwargs)
+    routes = (
+        list(create_agent_card_routes(agent_card=agent_card))
+        # 0.3 alias: buyer SDKs (e.g. @adcp/sdk) probe /.well-known/agent.json
+        # as a positive A2A signal. Same handler, no redirect round-trip.
+        + list(
+            create_agent_card_routes(
+                agent_card=agent_card, card_url="/.well-known/agent.json"
+            )
+        )
+        + list(create_jsonrpc_routes(**jsonrpc_kwargs))
     )
     app = Starlette(routes=routes)
 
--- a/src/adcp/server/auth.py
+++ b/src/adcp/server/auth.py
@@ -750,7 +750,8 @@ from a2a.utils.constants import (  # noqa: E402  (intentional placement after Be
 _A2A_DISCOVERY_PATHS: frozenset[str] = frozenset(
     {
         _A2A_AGENT_CARD_PATH,  # 1.0 canonical: ``/.well-known/agent-card.json``.
-        "/.well-known/agent.json",  # Legacy 0.3 alias retained by enable_v0_3_compat=True.
+        # Legacy 0.3 alias — route registered explicitly in create_a2a_server.
+        "/.well-known/agent.json",
     }
 )
 
--- a/src/adcp/server/auth.py
+++ b/src/adcp/server/auth.py
@@ -763,7 +763,8 @@ class A2ABearerAuthMiddleware:
     :func:`adcp.server.a2a_server.create_a2a_server` with this
     middleware to require a valid bearer header on every JSON-RPC
     request, while leaving the spec-mandated public discovery
-    surface (``/.well-known/agent-card.json``) accessible.
+    surface (``/.well-known/agent-card.json`` and the 0.3 alias
+    ``/.well-known/agent.json``) accessible.
 
     Designed to compose with a2a-sdk's
     :class:`DefaultServerCallContextBuilder`: on auth success the
--- a/tests/test_a2a_server.py
+++ b/tests/test_a2a_server.py
@@ -345,10 +345,13 @@ def test_create_a2a_server_creates_starlette_app():
     # Starlette app has .routes
     assert hasattr(app, "routes")
     route_paths = [r.path for r in app.routes]
-    # A2A well-known agent card endpoint
-    # 1.0 serves ``/.well-known/agent-card.json`` in addition to the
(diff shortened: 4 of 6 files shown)


bokelley:
## Pre-merge expert pass (fresh independent review)

- **code-reviewer:** APPROVE — no blockers. Verified the inverse-coverage test (`test_all_discovery_paths_are_registered_routes`) catches the bug class structurally; verified no shared-state risk from calling `create_agent_card_routes` twice (fresh closures, no globals). Nit only: both routes' `Route.name` defaults to `_get_agent_card`; `url_path_for` would resolve to whichever Starlette finds first. Nothing in SDK uses `url_path_for` for this — theoretical, non-blocking.
- **security-reviewer:** SAFE — no findings. Path was already auth-exempt; serving 200 instead of 404 doesn't expose anything new. Body is the static operator-authored agent card — no request-derived data, no per-tenant leakage. Exemption uses exact-set membership, not prefix match.
- **ad-tech-protocol-expert:** APPROVE — duplicate handler is the right call (308 would corrupt buyer SDKs that probe path-suffix as a transport-version signal). One forward-looking caveat: alias is registered unconditionally; today `enable_v0_3_compat=True` is hardcoded so this is a no-op concern, but if the flag becomes user-configurable the alias registration should gate on it. Worth a TODO, not a block.

Admin-merging.

bokelley:
Acknowledged — thanks for the independent expert pass and the forward-looking caveat on the `enable_v0_3_compat` flag. The TODO on gating alias registration is noted; if that flag becomes user-configurable a follow-up issue would be the right place to track it.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01TJqAJYiwTZoGQVdL1cEpDP

---

PR #627, opened 2026-05-10
feat(server): expose RequestContext.transport and current_transport ContextVar

### LINKED ISSUE #617: feat(server): expose RequestContext.transport (mcp | a2a)
## Problem

`RequestContext` doesn't carry the inbound transport identifier. Auth and tenant are exposed via ContextVars (`adcp.server.auth.current_principal`, etc.), but the request's transport is not.

Webhook payload shape selection is transport-dependent: A2A buyers expect `Task` / `TaskStatusUpdateEvent` envelopes; MCP buyers expect `McpWebhookPayload`. Platform methods that fire webhooks (or the in-house webhook service that does so on their behalf) need this signal — without it the platform defaults to one transport and the other side gets the wrong shape (we hit this as #202 locally).

## Workaround

We run an 85-LOC ASGI middleware (`TransportDetectMiddleware`) that classifies the URL path (`/mcp` or `/mcp/*` → MCP; host root including `/.well-known/agent-card.json` → A2A; admin paths → no-op) and stuffs the result into a `current_transport` ContextVar that platform code reads. Works, but it's URL-path inference outside the SDK's dispatch, and any change to how the framework routes transports silently breaks our detection.

## Proposed SDK shape

```python
from adcp.server import current_transport  # contextvar set by the dispatcher

# Or as a field on RequestContext:
@dataclass
class RequestContext:
    ...
    transport: Literal["mcp", "a2a"]
```

The dispatcher already knows which transport invoked it (it picked the route). It's just not surfacing the value to handlers.

## Files

- Our workaround: [core/middleware/transport_detect.py](https://github.com/prebid/salesagent/blob/main/core/middleware/transport_detect.py)
- Reader of the ContextVar: `core/platforms/_delegate._build_identity` — used to pick A2A vs MCP webhook payload shape.
- Local context: this is the same need that surfaced as #202 (close); a public `transport` field would have made #202 trivial.

bokelley:
Closes #617

## Summary

Adds `RequestContext.transport: Literal["mcp", "a2a"] | None` so handler methods can branch on the inbound wire protocol without 85 lines of ASGI URL-inference middleware. Also exports `current_transport: ContextVar[Literal["mcp", "a2a"] | None]` from `adcp.server` for webhook services and other code running outside a handler call stack. Fixes a pre-existing gap where `current_principal` and `current_principal_metadata` were only accessible via the private `adcp.server.auth` path rather than the public `adcp.server` surface.

```python
# In a handler method:
async def create_media_buy(self, params, context):
    if context.transport == "a2a":
        await send_task_status_update(...)
    else:
        await send_mcp_webhook(...)

# In a webhook service outside the handler call stack:
from adcp.server import current_transport
transport = current_transport.get()  # "mcp" | "a2a" | None
```

## What was tested

```
ruff check src/                       # All checks passed
mypy src/adcp/                        # No errors
pytest tests/test_decisioning_dispatch.py tests/test_auth_middleware.py -v
# 93 passed, 13 warnings
pytest tests/ -q                      # 4213 passed, 1 pre-existing TLS network test failed (unrelated), 33 skipped
```

## Pre-PR review

- **code-reviewer:** approved — no CI-blocking findings. Flagged (a) validation of raw metadata value, (b) SDK-internal keys leaking into handler-visible `ctx.metadata`, (c) missing ContextVar assertion in test. All three fixed before PR open.
- **dx-expert:** approved — blocker was `current_transport.set()` without reset token; resolved via explanatory comment noting asyncio task-context isolation (each request task copies context at creation, so `set()` is task-scoped and cannot bleed across requests; resetting immediately after `_build_request_context` returns would undo the value before any handler code runs). Also fixed: cross-link to `current_transport` in field docstring, note that custom `context_factory` omitting `metadata["transport"]` also produces `None`.

**Nits surfaced, not fixed:**
- `current_tenant` (from `tenant_router.py`) and `current_principal` (from `auth.py`) are heterogeneous in origin and type in `__init__.py` — pre-existing, noted for a future cleanup PR.
- `@pytest.mark.parametrize` is now used for the transport extraction test (clearer failure messages vs. the original `for` loop).

---

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01UifYgpi26gbfXmhrNRFDvK

---
_Generated by [Claude Code](https://claude.ai/code/session_01UifYgpi26gbfXmhrNRFDvK)_

### DIFF
--- a/src/adcp/decisioning/context.py
+++ b/src/adcp/decisioning/context.py
@@ -18,7 +18,7 @@ import warnings
 from collections.abc import Awaitable, Callable, Mapping
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
-from typing import TYPE_CHECKING, Any, ClassVar, Generic
+from typing import TYPE_CHECKING, Any, ClassVar, Generic, Literal
 
 from typing_extensions import TypeVar
 
@@ -533,6 +533,14 @@ class RequestContext(ToolContext, Generic[TMeta]):
     * Idempotency scope? → don't touch; the framework owns this.
     * Logging request provenance? → log all four; they're cheap.
 
+    :param transport: The wire protocol that dispatched this call —
+        ``"mcp"`` or ``"a2a"``. ``None`` only when ``RequestContext``
+        is constructed directly in tests without a transport-aware
+        ``ToolContext`` (production dispatch always populates this
+        field). Note: even when the server is started with
+        ``transport="both"``, individual requests always resolve to
+        exactly one of ``"mcp"`` or ``"a2a"`` — this field never
+        carries ``"both"``.
     :param state: Sync reads of framework-owned in-flight workflow
         state. Default is :class:`adcp.decisioning.state._NotYetWiredStateReader`
         — returns empty values + emits one-time UserWarning per
@@ -560,6 +568,7 @@ class RequestContext(ToolContext, Generic[TMeta]):
     auth_info: AuthInfo | None = None
     auth_principal: str | None = None
     buyer_agent: BuyerAgent | None = None
+    transport: Literal["mcp", "a2a"] | None = None
     now: datetime = field(default_factory=lambda: datetime.now(timezone.utc))
     state: StateReader = field(default_factory=_make_default_state_reader)
     resolve: ResourceResolver = field(default_factory=_make_default_resolver)
--- a/src/adcp/decisioning/context.py
+++ b/src/adcp/decisioning/context.py
@@ -534,13 +534,15 @@ class RequestContext(ToolContext, Generic[TMeta]):
     * Logging request provenance? → log all four; they're cheap.
 
     :param transport: The wire protocol that dispatched this call —
-        ``"mcp"`` or ``"a2a"``. ``None`` only when ``RequestContext``
-        is constructed directly in tests without a transport-aware
-        ``ToolContext`` (production dispatch always populates this
-        field). Note: even when the server is started with
+        ``"mcp"`` or ``"a2a"``. ``None`` when ``RequestContext`` is
+        constructed in tests without a transport-aware ``ToolContext``,
+        or when a custom ``context_factory`` omits
+        ``metadata["transport"]``. Production dispatch always populates
+        this field. Note: even when the server is started with
         ``transport="both"``, individual requests always resolve to
         exactly one of ``"mcp"`` or ``"a2a"`` — this field never
-        carries ``"both"``.
+        carries ``"both"``. For code running outside a handler call
+        stack, read :data:`adcp.server.current_transport` instead.
     :param state: Sync reads of framework-owned in-flight workflow
         state. Default is :class:`adcp.decisioning.state._NotYetWiredStateReader`
         — returns empty values + emits one-time UserWarning per
--- a/src/adcp/decisioning/dispatch.py
+++ b/src/adcp/decisioning/dispatch.py
@@ -1059,6 +1059,7 @@ def _build_request_context(
     # Local import keeps the layering local — read the bearer ContextVar
     # without forcing a top-level dep on adcp.server.auth.
     from adcp.server.auth import current_principal as _current_principal
+    from adcp.server.auth import current_transport as _current_transport
 
     if auth_info is None:
         bearer_principal = _current_principal.get()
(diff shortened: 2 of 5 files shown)


bokelley:
## Adopter confirmation (filed #617)

Read the diff. **This kills our [core/middleware/transport_detect.py](https://github.com/bokelley/salesagent/blob/main/core/middleware/transport_detect.py) cleanly** (85 LOC of URL-path classification + ContextVar plumbing).

Two specific things I like:

1. **Both surfaces exposed:** \`ctx.transport\` for handler-scope reads, \`adcp.server.current_transport\` ContextVar for code outside the call stack (our webhook service was the original need — see prebid/salesagent#202). Either works in the right context; neither is forced.
2. **Stripping \`metadata["transport"]\` from handler-visible metadata** — the right call. Adopters reading the typed field can't accidentally also read an undocumented dict path that might drift.

### One verification request

The PR reads from \`tool_ctx.metadata["transport"]\`. This depends on the SDK's MCP and A2A dispatchers always populating that key on the \`ToolContext\` they hand to \`_build_request_context\`. The diff covers \`_build_request_context\` cleanly but I don't see the dispatcher-side population in the diff scope. Can you confirm both paths set \`metadata["transport"]\` before this lands? An adopter who hits a \`transport=None\` in production handler code would have a confusing time.

If the dispatcher wiring is in another PR, please link it; if it's already on \`main\`, the test count (15 passing) suggests it's fine.

### Adoption plan

Once this lands and we bump:
1. Replace \`from core.middleware.transport_detect import current_transport\` with \`from adcp.server import current_transport\` in our webhook service.
2. Convert \`platform._build_identity\` to read \`ctx.transport\` directly instead of the ContextVar.
3. Delete \`core/middleware/transport_detect.py\` and remove the entry from \`_serve_kwargs\`.

Small, mechanical change on our side. Ready when you are.

bokelley:
@bokelley — confirmed, no separate PR needed. The dispatcher-side wiring was already on `main` before this diff, in two places:

1. **Default `context_factory`** (`src/adcp/server/serve.py:487`) — the factory `serve()` uses when no custom one is supplied:
   ```python
   metadata={"tool_name": meta.tool_name, "transport": meta.transport}
   ```

2. **`auth_context_factory`** (`src/adcp/server/auth.py:420-423`) — the auth-aware factory used when `AuthConfig` is provided:
   ```python
   combined_metadata: dict[str, Any] = {
       **principal_metadata,
       "transport": meta.transport,
       ...
   }
   ```

Both receive `RequestMetadata` where `transport: Literal["mcp", "a2a"]` is a required, non-optional field populated by the server framework itself for every MCP and A2A call — not by user code. The 15 passing transport-extraction tests are a reliable signal.

The only `None` path (documented in the `RequestContext.transport` docstring added by this PR) is a custom `context_factory` that doesn't forward `meta.transport` into `metadata`. That's a user-owned factory, not an SDK dispatch path, so production SDK dispatch is always populated.

---

bokelley:
## Verified dispatcher write sites

Confirmed `metadata["transport"]` is populated everywhere a `RequestMetadata` is constructed:

- **`auth_context_factory`** (the canonical SDK-provided factory at `src/adcp/server/auth.py` consumed in `src/adcp/server/serve.py:487`):
  ```python
  metadata={"tool_name": meta.tool_name, "transport": meta.transport}
  ```
- **MCP path** when `context_factory=` is wired (`src/adcp/server/serve.py:2049`):
  ```python
  meta = RequestMetadata(tool_name=name, transport="mcp")
  ```
- **A2A path** when `context_factory=` is wired (`src/adcp/server/a2a_server.py:294-298`):
  ```python
  meta = RequestMetadata(tool_name=skill_name, transport="a2a", request_id=request.task_id)
  ```

### One caveat to call out — already documented in the field docstring

Adopters who **do not wire a `context_factory`** fall through to bare-`ToolContext` construction (`ToolContext()` in `mcp_tools.call_tool`, `_tool_context_from_request(request)` on the A2A side at `a2a_server.py:539`). Neither populates `metadata["transport"]`, so `ctx.transport` resolves to `None`.

The PR's docstring on `RequestContext.transport` already calls this out:

> ``None`` when ``RequestContext`` is constructed in tests without a transport-aware ``ToolContext``, or when a custom ``context_factory`` omits ``metadata["transport"]``. Production dispatch always populates this field.

The "production dispatch always populates" claim is true for **adopters using the SDK-provided `auth_context_factory` or any factory that propagates `meta.transport`** — which is the documented pattern. Custom factories that omit it get `None`; that's surfaced as a documented contract rather than a silent footgun.

Net: the read site (`_build_request_context`) is correctly paired with write sites at every dispatcher entry point. ✅

bokelley:
Thanks for the write-site audit, @bokelley. The `None`-on-missing-factory contract being documented in the field docstring rather than enforced with a runtime error is the right call — custom factories are a supported escape hatch and the docstring makes the gap explicit. No follow-up action needed from this side.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_01QMJzJHYW6SzcPUZTAE9WTU_

---

PR #628, opened 2026-05-10
feat(server): add TenantRegistry with per-tenant health tracking

### LINKED ISSUE #619: feat(server): TenantRegistry parity with JS createTenantRegistry
## Problem

The Python SDK ships `CallableSubdomainTenantRouter` (host-routing callback, PR #544) and `LazyPlatformRouter` (per-tenant platform factory, PR #547). The JS SDK ships `createTenantRegistry` which is a higher-level primitive built on top of the same building blocks. Adopters running multi-tenant Python deployments (we're the most advanced one) end up reinventing the registry layer.

## What JS has that Python doesn't

From `@adcp/sdk/server`'s `createTenantRegistry` (see [adcp-client/skills/build-decisioning-platform/advanced/MULTI-TENANT.md](https://github.com/adcontextprotocol/adcp-client/blob/main/skills/build-decisioning-platform/advanced/MULTI-TENANT.md)):

1. **Per-tenant health states** — `pending` (registered, not yet validated, refused with 503), `healthy` (serving), `unverified` (was healthy, transient validation failure, graceful-degrade), `disabled` (permanent failure, refused until admin `recheck()`). Per-tenant — one bad tenant doesn't block others.
2. **Runtime `register(tenantId, config)` / `unregister(tenantId)`** — add/remove tenants without restarting the process. Admin webhook saves a new tenant row, calls register, the next request to that host resolves it. We need this; today we plumb it ourselves through the admin flow.
3. **`recheck(tenantId)`** — re-validate a tenant after key rotation or config change. Status transitions `disabled → healthy` without a traffic gap.
4. **`awaitFirstValidation: true`** — boot-time semantic where `register()` doesn't return until the tenant has been validated, so the first request after register doesn't race the validation roundtrip.
5. **`resolveByHost(host)`** — synchronous lookup returning the registered server (or null). Composes naturally with `serve()`.

JS-specific bits we don't need: JWKS validation (we use principal-token bearer auth, not JWT). The registry should be JWKS-agnostic — adopters that want JWKS can pass a validator, adopters that don't pass nothing.

## What we have today

`core/main.py` builds:
- A `CallableSubdomainTenantRouter` with a 60-second TTL cache (host → Tenant lookup against our DB)
- A `LazyPlatformRouter` with per-tenant `DecisioningPlatform` factory
- An ad-hoc admin-flow `invalidate(host)` call when a tenant is created / deactivated / has its subdomain rotated

What's missing relative to JS:
- No health states. A misconfigured tenant (e.g. GAM credentials missing) only fails on first request, with a generic 500 — there's no \"this tenant is disabled\" classifier the LB or admin UI can observe.
- No runtime register without invalidate-the-cache plumbing.
- No `recheck()` — config changes require either a process restart or manual cache invalidation.
- No first-validation boot semantic — the first request to a freshly-registered tenant pays the platform-build cost.

## Proposed SDK shape

```python
from adcp.server import TenantRegistry, BearerTokenAuth, serve

registry = TenantRegistry(
    default_serve_options={
        \"name\": \"my-multi-tenant-host\",
        \"validation\": {\"requests\": \"strict\", \"responses\": \"strict\"},
    },
    # Optional validator — adopters using JWT can pass a JWKS validator;
    # principal-token adopters pass None.
    validator=None,
    auto_validate=True,
)

# Register at boot
for tenant in load_tenants_from_db():
    await registry.register(
        tenant.id,
        agent_url=tenant.agent_url,
        platform=build_platform_for(tenant),
        await_first_validation=True,
    )

# Resolve per request
def resolve(ctx) -> AdcpServer:
    resolved = registry.resolve_by_host(ctx.host)
    if resolved is None or resolved.health == \"disabled\":
        raise HTTPException(503)
    return resolved.server

serve(resolve, auth=BearerTokenAuth(validate_token=...), port=os.environ[\"PORT\"])

# Runtime admin operations
await registry.register(tenant_id, agent_url=..., platform=...)  # add
registry.unregister(tenant_id)                                    # remove
await registry.recheck(tenant_id)                                 # re-validate
status = registry.health(tenant_id)                               # observe
```

The internals reuse `CallableSubdomainTenantRouter` and `LazyPlatformRouter` — `TenantRegistry` is the higher-level primitive that composes them with health tracking and runtime mutation, matching the JS surface area.

## Why this matters

- Multi-tenant SaaS deployments are the common shape for AdCP sellers. Every Python adopter at our scale will land here.
- Building it ourselves means each adopter has slightly-different health semantics, slightly-different admin webhook plumbing, slightly-different cache-invalidation rules. A canonical primitive collapses that variation.
- Closes the JS↔Python parity gap on the most-touched server-side primitive.

## Files

- Our DIY implementation: [core/main.py](https://github.com/prebid/salesagent/blob/main/core/main.py) (`_resolve_tenant`, `build_subdomain_router`, `build_platform_for_tenant`, `build_router`)
- JS reference: `@adcp/sdk/server`'s [tenant-registry.ts](https://github.com/adcontextprotocol/adcp-client/blob/main/src/lib/server/decisioning/tenant-registry.ts) and [tenant-store.ts](https://github.com/adcontextprotocol/adcp-client/blob/main/src/lib/server/decisioning/tenant-store.ts).

bokelley:
Closes #619

Adds `TenantRegistry` — a higher-level multi-tenant primitive that closes the JS↔Python parity gap on `createTenantRegistry`. Adopters pre-build per-tenant `DecisioningPlatform` instances and register them; the registry tracks health state and surfaces a synchronous `resolve_by_host` for the hot request path.

## Summary

- New `src/adcp/server/tenant_registry.py` exporting `TenantRegistry`, `TenantResolution`, `TenantHealthState`, `TenantValidator`
- `TenantRegistry` manages four health states per tenant: `pending → healthy / disabled`, `healthy → unverified` (graceful-degrade on failed recheck), `unverified / disabled → healthy` (successful recheck)
- Per-tenant `asyncio.Lock` prevents TOCTOU races on concurrent `recheck()` calls; post-validator guard prevents zombie `_health` entries when `unregister()` races with an in-flight `recheck()`
- `resolve_by_host` is synchronous (in-memory dict updated eagerly by `register`/`unregister`) — intentional departure from the JS async variant; the Python registry owns its mapping directly
- `validator=None` is principal-token mode (always healthy); JWKS or custom health-check adopters pass a `(tenant_id, agent_url) -> bool` callable (sync or async)
- `auto_validate` kwarg from the issue proposal dropped per DX review; validator presence is the opt-in

**Nits noted (not fixed in this PR):**
- `_normalize_host` does not handle bare bracketless IPv6 hosts (edge case, DNS names only in practice)
- `health()` returning `None` for unknown tenants vs `platform_for_tenant` raising `KeyError` — intentional for a probe/observation method; documented in docstring

## What-tested

- `ruff check` — clean
- `mypy src/adcp/server/tenant_registry.py --ignore-missing-imports` — no errors in new file
- `pytest tests/test_tenant_registry.py -v` — **25/25 passed**
  - Lifecycle: register, unregister, resolve_by_host
  - Health state machine: all 4 states, all transition arms
  - Host normalization: port stripping, case folding, URL change
  - Validators: sync, async, raising
  - Concurrency: concurrent rechecks (lock serializes), unregister-during-recheck (no zombie entry)

## Pre-PR review

- **code-reviewer:** approved — ruff F401/I001 nits (both fixed before commit); zombie health entry blocker (fixed); `_platforms` type tightened to `dict[str, DecisioningPlatform]`; `serve_options` returns copy (fixed)
- **dx-expert:** approved (1 nit noted) — `serve_options` public property added; docstring usage snippet added

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_019ucMgF6YS9X5bygYUADztx

---
_Generated by [Claude Code](https://claude.ai/code/session_019ucMgF6YS9X5bygYUADztx)_

### DIFF
--- a/src/adcp/server/__init__.py
+++ b/src/adcp/server/__init__.py
@@ -145,6 +145,7 @@ from adcp.server.serve import (
 )
 from adcp.server.sponsored_intelligence import SponsoredIntelligenceHandler
 from adcp.server.tenant_registry import (
+    PlatformFactory,
     TenantHealthState,
     TenantRegistry,
     TenantResolution,
@@ -224,6 +225,7 @@ __all__ = [
     "IdempotencyStore",
     "MemoryBackend",
     # Multi-tenant registry with health tracking
+    "PlatformFactory",
     "TenantHealthState",
     "TenantRegistry",
     "TenantResolution",
--- a/src/adcp/server/__init__.py
+++ b/src/adcp/server/__init__.py
@@ -144,6 +144,12 @@ from adcp.server.serve import (
     serve,
 )
 from adcp.server.sponsored_intelligence import SponsoredIntelligenceHandler
+from adcp.server.tenant_registry import (
+    TenantHealthState,
+    TenantRegistry,
+    TenantResolution,
+    TenantValidator,
+)
 from adcp.server.tenant_router import (
     CallableSubdomainTenantRouter,
     InMemorySubdomainTenantRouter,
@@ -217,6 +223,11 @@ __all__ = [
     # Idempotency middleware (AdCP #2315 seller side)
     "IdempotencyStore",
     "MemoryBackend",
+    # Multi-tenant registry with health tracking
+    "TenantHealthState",
+    "TenantRegistry",
+    "TenantResolution",
+    "TenantValidator",
     # Subdomain tenant routing
     "CallableSubdomainTenantRouter",
     "InMemorySubdomainTenantRouter",
--- a/src/adcp/server/tenant_registry.py
+++ b/src/adcp/server/tenant_registry.py
@@ -552,6 +552,9 @@ class TenantRegistry:
                 )
                 if tenant_id in self._health:
                     self._health[tenant_id] = "disabled"
+                    # Drop the factory so subsequent resolve() calls don't re-invoke
+                    # it — a disabled tenant needs operator intervention via recheck().
+                    self._factories.pop(tenant_id, None)
                 return None
 
             try:
@@ -565,6 +568,7 @@ class TenantRegistry:
                 )
                 if tenant_id in self._health:
                     self._health[tenant_id] = "disabled"
+                    self._factories.pop(tenant_id, None)
                 return None
 
             # Guard: unregister() may have run while we awaited factory/validator.
@@ -574,11 +578,14 @@ class TenantRegistry:
             if ok:
                 self._platforms[tenant_id] = platform
                 self._health[tenant_id] = "healthy"
+                # Factory no longer needed — platform is cached in _platforms.
+                self._factories.pop(tenant_id, None)
                 return TenantResolution(
                     tenant_id=tenant_id, health="healthy", platform=platform
                 )
             else:
                 self._health[tenant_id] = "disabled"
+                self._factories.pop(tenant_id, None)
                 return None
 
     @property
--- a/src/adcp/server/tenant_registry.py
+++ b/src/adcp/server/tenant_registry.py
@@ -1,0 +1,396 @@
+"""TenantRegistry — higher-level multi-tenant management primitive.
+
+Provides JS ``createTenantRegistry`` parity for Python multi-tenant
+deployments. Composes per-tenant health tracking with runtime
+register/unregister/recheck, closing the JS↔Python parity gap on the
+most-touched server-side primitive.
+
+Adopters pre-build per-tenant :class:`~adcp.decisioning.DecisioningPlatform`
+instances and register them here. The registry tracks health state and
+surfaces :meth:`TenantRegistry.resolve_by_host` for the request path.
+
+Comparison with lower-level building blocks:
+
+* :class:`~adcp.server.CallableSubdomainTenantRouter` — host→Tenant lookup
+  with TTL cache. Suitable when tenant routing is all you need.
+* :class:`~adcp.decisioning.LazyPlatformRouter` — per-tenant
+  :class:`~adcp.decisioning.DecisioningPlatform` factory with LRU+TTL
+  cache. Suitable when you need lazy platform construction.
+* :class:`TenantRegistry` — combines health tracking + runtime mutation
+  (register/unregister/recheck) into one object. Reach for this when
+  multi-tenant SaaS topology requires observability into per-tenant
+  health state and runtime admin operations without a process restart.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import inspect
+import logging
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, Literal
+from urllib.parse import urlparse
+
+if TYPE_CHECKING:
+    from adcp.decisioning.platform import DecisioningPlatform
+
+logger = logging.getLogger(__name__)
+
+#: Per-tenant health state. See :class:`TenantRegistry` for semantics.
+TenantHealthState = Literal["pending", "healthy", "unverified", "disabled"]
+
+#: Validator callable. Takes ``(tenant_id, agent_url)`` and returns
+#: ``True`` when the tenant is valid, ``False`` otherwise.
+#: May be sync or async — the registry awaits at call time.
+TenantValidator = Callable[[str, str], "bool | Awaitable[bool]"]
+
+
+@dataclass(frozen=True)
+class TenantResolution:
+    """Result of :meth:`TenantRegistry.resolve_by_host`.
+
+    :param tenant_id: Stable identifier for the resolved tenant.
+    :param health: Current health state. Callers gate traffic on this —
+        typically 503 for ``pending`` and ``disabled``, serve for
+        ``healthy`` and ``unverified``.
+    :param platform: The :class:`~adcp.decisioning.DecisioningPlatform`
+        for this tenant. Pass to :func:`adcp.decisioning.serve` or use
+        with a :class:`~adcp.decisioning.PlatformRouter`.
+    """
+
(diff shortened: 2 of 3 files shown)


bokelley:
## Adopter feedback (filed #619 — most-advanced Python multi-tenant adopter)

Read the full diff. The health-state machine, validator hook, and runtime mutation surface match what we asked for and look great. **One critical design issue blocks adoption for us as written: the registration model is eager, and we need lazy.**

### Eager-vs-lazy mismatch

The PR's docstring example:

\`\`\`python
for tenant in load_tenants_from_db():
    await registry.register(
        tenant.id,
        agent_url=tenant.agent_url,
        platform=build_platform_for(tenant),  # ← built at boot
    )
\`\`\`

We currently use \`LazyPlatformRouter\` *precisely because we don't want to pay per-tenant platform-build cost at boot*. \`build_platform_for_tenant\` for a GAM tenant means a network handshake to the GAM auth server, credential fetch from KMS, and inventory-manager construction. Multiplied across N tenants, boot time scales linearly with tenant count. We do this lazily on first request per tenant ([core/main.py:191](https://github.com/bokelley/salesagent/blob/main/core/main.py#L191)).

If we adopt \`TenantRegistry\` as written, we'd either:
1. Pay the boot-time cost (regression), or
2. Skip \`register()\` and lazily call it from a \`resolve_by_host\` shim — defeating the point of the registry, since unregistered tenants 503

### Proposed fix: lazy-factory variant

Add a \`register_lazy()\` (or \`factory=\` kwarg on \`register\`) that takes a builder callable instead of an already-built platform:

\`\`\`python
async def register_lazy(
    self,
    tenant_id: str,
    *,
    agent_url: str,
    factory: Callable[[str], Awaitable[DecisioningPlatform]],
    await_first_validation: bool = False,
) -> None:
    """Register a tenant with a lazy platform factory.

    The platform is built on first resolve_by_host hit, then cached.
    Subsequent resolves return the cached instance. Suitable for
    deployments with many tenants where eager construction is too
    expensive.
    """
\`\`\`

Internal: the registry holds the factory + a \`platform: DecisioningPlatform | None\` slot. \`resolve_by_host\` checks the slot, awaits the factory if empty, fills the slot. The validator + health-state machine work unchanged on the resolved platform.

This composes cleanly with the existing eager API — adopters with few tenants pre-build, adopters with many lazy-build, same registry primitive.

### Health states + lazy: behavioral note

With lazy registration, \`pending\` would mean "registered, factory not yet invoked." The first \`resolve_by_host\` triggers factory + validator. Health transitions \`pending → healthy\` on success, \`pending → disabled\` on factory failure. Matches the eager semantics.

### Other observations (non-blocking)

- \`_normalize_host\` strips port — correct for our setup, but note that some load balancers re-emit non-default ports in \`X-Forwarded-Host\`. Worth a docstring note.
- The \`default_serve_options\` plumbing is nice but the docstring example \`serve(platform, **registry.serve_options)\` only works for single-tenant deployments. Multi-tenant deployments don't pass a single platform to \`serve()\` — they pass a router. Worth clarifying or removing.
- TOCTOU lock per-tenant is correct. Document that locks aren't garbage-collected on \`unregister\` (the current code preserves them, which is right — race-resistant — but adopters might wonder).

### Bottom line

If the lazy-factory variant lands, we adopt this PR and delete our hand-rolled tenant lifecycle. As written, we'd have to skip it or wrap it. The lazy mode is the difference between "useful primitive for single-host SaaS" and "useful primitive for multi-tenant SaaS at any scale." JS \`createTenantRegistry\` has the same eager-only limitation — Python could ship the better surface here.

bokelley:
## Follow-up: lazy-factory variant landed

Three commits on this branch address your feedback:

**`register_lazy()` + `async resolve()`** — `register_lazy(tenant_id, *, agent_url, factory, await_first_validation=False)` takes a `PlatformFactory = Callable[[str], Awaitable[DecisioningPlatform]]`; the platform is built on the first `await registry.resolve(host)` hit and cached. Concurrent first-hits serialize on the per-tenant lock — one factory call per tenant. Existing `resolve_by_host` (sync) is unchanged and returns `None` until the platform is built.

**Health semantics** — `pending` = registered, factory not yet invoked. First `resolve()`: `pending → healthy` on success, `pending → disabled` on factory/validator failure. Factory is cleared on `disabled` so subsequent `resolve()` calls return `None` without retrying; recovery requires `register_lazy()` + `recheck()`.

**Two correctness issues found + fixed in pre-PR review:**
- `register_lazy(await_first_validation=True)` + validator returns `False`: platform was being cached in `_platforms` and the factory was leaking into `_factories`. Fixed to match `resolve()` cold-path: discard the rejected platform, clear the factory.
- `resolve()` docstring implied it always returns `None` when the validator rejects — only true on the cold-path. Fast-path (eager or previously-resolved lazy) returns `TenantResolution(health="disabled")`. Docstring and inline warning updated: always check `result.health`, never gate solely on `result is None`.

**Your non-blocking observations** all addressed: `_normalize_host` load-balancer port caveat added, `serve_options` single-vs-multi-tenant clarification updated, lock lifecycle documented.

**One nit to flag:** `async def resolve(host)` shares its name and arity with `SubdomainTenantRouter.resolve` — a `@runtime_checkable Protocol` in this package. Mypy catches the type mismatch; duck-typing doesn't. Added a class-docstring warning. Happy to rename to `resolve_platform` if you'd prefer to close the pitfall before this merges.

41 tests pass.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01DRv6qahN7Jjt3Q4oxGBXkd

---

Issue #623, opened 2026-05-10
Typed dispatcher rejects valid request when `account` is omitted on auth-resolved tools


bokelley:
## Problem

The library's typed dispatcher rejects requests on tools where the spec declares \`account\` optional — specifically when the seller resolves identity from the auth chain and the buyer doesn't send a top-level \`account\` ref.

Affected wire tools (3.0.6+):
- \`create_media_buy\`
- \`sync_creatives\`
- \`activate_signal\`
- (others sharing the AccountReference-bound request shape)

Per spec, when an adopter is in implicit-resolution mode (auth principal IS the account scope) the buyer can omit \`account\`. The dispatcher's pre-handler validation insists on the field being present, blocking the request before the adopter's impl can resolve from \`AuthInfo\`.

## Workaround on the adopter side

Inject a sentinel \`account: {account_id: \"auth-chain\"}\` placeholder at the wire boundary so the validator passes; the impl ignores the placeholder and reads identity from \`ResolvedIdentity\`. [bokelley/salesagent's \`SpecDefaultsMiddleware\` does this](https://github.com/bokelley/salesagent/blob/main/core/middleware/spec_defaults.py) for the four tools listed above. Workable but every adopter has to ship the equivalent.

## Suggested fix

Make \`account\` optional on the typed dispatcher when the surface declares implicit-resolution support. Concretely:

1. \`AccountStore.resolution\` is \`'derived' | 'implicit' | 'explicit'\` — the dispatcher already has this signal at boot.
2. For tools that accept \`account\` (\`create_media_buy\`, \`sync_creatives\`, \`activate_signal\`, etc.), skip the required-field check on \`account\` when the wired \`AccountStore.resolution != 'explicit'\`.
3. Keep the field required for explicit-mode adopters (where \`account.account_id\` IS the resolution input).

This closes the gap without changing the wire schema (which already documents \`account\` as optional in implicit/derived mode).

## Related

- bokelley/salesagent's middleware workaround: \`core/middleware/spec_defaults.py\` \`_AUTH_FILLED_TOOLS\` set.
- The \`auth-chain\` placeholder string is contained (impl reads identity from elsewhere) but is fragile — any adopter forgetting the middleware silently gets \`INVALID_REQUEST\` for buyers in implicit mode.

bokelley:
## Triage

**Classification:** Bug
**Bucket(s):** handlers (label not found in repo — flagged in run summary)
**Status:** drafting-pr
[...]

PR #629, opened 2026-05-10
feat(server): pre-validation request hook for spec-default injection (#614)

### LINKED ISSUE #614: feat(server): pre-validation request hook / spec-default registry
## Problem

Some 4.4+ schemas mark fields as `required` at the wire level even though the spec instructs sellers to apply a default for missing values from pre-v3 clients. Examples:

- `GetProductsRequest.buying_mode` — spec text: *"Sellers receiving requests from pre-v3 clients without buying_mode SHOULD default to 'brief'."*
- `account` and `idempotency_key` on tools where the seller resolves identity from the auth chain (we backfill `account_id="auth-chain"` and `idem-{uuid4}` placeholders so strict validation passes; the impl ignores them)
- `format_id` shape: 4.4 made it a structured `{agent_url, id}` reference; pre-4.4 buyers send a bare string
- `assets[].asset_type` on `sync_creatives` — buyers send `{"image": {...}}` without `asset_type`, which makes the discriminated union unable to pick a branch
- Image asset variant strictly requires `width`/`height`; pre-4.4 buyers often send image URLs without dims (we demote `image` → `url`)

The typed dispatcher validates payloads against the library Pydantic models **before** the platform handler runs, so a per-handler `model_validator` cannot apply spec-mandated defaults in time.

## Workaround (the most advanced Python adopter)

We ship a 273-LOC ASGI middleware (`SpecDefaultsMiddleware`) that bytes-rewrites JSON-RPC `tools/call` bodies and the matching A2A skill payloads, applying defaults in-place before the SDK validator runs. It works, but:

1. Every adopter who hits AdCP 4.4+ strictness needs the same set of defaults.
2. The middleware sits *outside* the SDK's validation boundary by definition.
3. Bytes-level body rewriting is brittle — adding a new defaulted field means another hand-rolled patcher.

## Proposed SDK shape

Either of:

**A) Pre-validation request hook (per-tool or global):**
```python
serve(
    router,
    pre_validation_hooks={
        "get_products": lambda args: {**args, "buying_mode": args.get("buying_mode", "brief")},
        # or global hook receiving (tool_name, args)
    },
)
```

**B) Declarative spec-default registry:**
```python
from adcp.server import SpecDefaults

serve(
    router,
    spec_defaults=SpecDefaults(
        get_products={"buying_mode": "brief"},
        sync_creatives={"asset_type_inference": True, "format_id_normalize": True},
    ),
)
```

Either approach kills our 273-LOC middleware and gives every Python adopter the same spec-conformant defaulting behaviour.

## Files

- Our workaround: [core/middleware/spec_defaults.py](https://github.com/prebid/salesagent/blob/main/core/middleware/spec_defaults.py)
- Per-spec-text justification for each default is in the module docstring + per-rule comments.

bokelley:
Closes #614

## What this adds

`pre_validation_hooks: dict[str, Callable[[str, dict], dict]] | None` on every public entry point — `serve()`, `create_mcp_server()`, `create_a2a_server()`, `MCPToolSet`, `create_mcp_tools()`, and `create_tool_caller()`.

The per-tool hook runs on the raw wire dict **before** `validate_request` (JSON schema) and `model_validate` (Pydantic), so it can supply spec-mandated defaults for pre-v3 buyers that omit required fields. This eliminates the 273-line ASGI middleware workaround the issue author ships today.

**Signature:** `(tool_name: str, raw_args: dict) -> dict`

**Quick example:**

```python
serve(
    router,
    pre_validation_hooks={
        "get_products": lambda n, a: {**a, "buying_mode": a.get("buying_mode", "brief")},
    },
)
```

**Behaviour guarantees:**
- `None` default → zero overhead on the hot path (no branch entered)
- Hook exceptions → `ADCPTaskError(code="INVALID_REQUEST")` — never `INTERNAL_ERROR`
- `raw_params` snapshotted **before** hook call → context echo always reflects original wire input, not server-injected defaults
- Per-tool scoping: hooks dict is fanned out to individual `create_tool_caller()` closures at registration time, not at call time

## Files changed

| File | Change |
|---|---|
| `src/adcp/server/mcp_tools.py` | Add `pre_validation_hook` param to `create_tool_caller()`, `MCPToolSet`, `create_mcp_tools()` |
| `src/adcp/server/serve.py` | Thread `pre_validation_hooks` through `ServeConfig`, `serve()`, `_serve_mcp/a2a`, `create_mcp_server()`, `_register_handler_tools()` |
| `src/adcp/server/a2a_server.py` | Thread through `ADCPAgentExecutor.__init__()` and `create_a2a_server()` |
| `src/adcp/decisioning/serve.py` | Add param + docstring example to decisioning `serve()` |
| `tests/test_pre_validation_hooks.py` | New — 7 tests covering all key invariants |

## Tested

- `pytest tests/ -q` → **4161 passed, 0 new failures** (pre-existing `test_real_tls_handshake_still_validates_hostname` excluded — fails on main too)
- `mypy src/adcp/` → **Success: no issues found in 783 source files**
- `ruff check src/` → **All checks passed**

## Pre-PR expert review

Two expert agents (code-reviewer + dx-expert) reviewed the diff independently. One blocker was found and fixed before this PR was opened:

**BLOCKER (fixed):** `raw_params = params` was originally assigned *after* the hook ran, meaning server-injected defaults (e.g. `"buying_mode": "brief"`) would be echoed back to the buyer as if they sent them — a direct violation of the AdCP context-echo contract. Fixed by moving the snapshot to *before* the hook call. The fix is covered by `test_hook_does_not_pollute_context_echo`, which sends a wire payload with a `context` field, uses a stripping hook that returns a completely new dict (no `context` key), and asserts the response still echoes the original wire context.

**NITs (not fixed, left for reviewer discretion):**
- `ServeConfig` and `a2a_server.py` use `dict[str, Callable[..., Any]]`; `mcp_tools.py`/`MCPToolSet` use the precise `dict[str, Callable[[str, dict[str, Any]], dict[str, Any]]]` — the looser type at outer layers loses IDE completions
- No `PreValidationHook` type alias exported from `adcp.server` (compare: `SkillMiddleware` is exported); adopters writing typed code must spell out the full callable signature

---

*This PR was opened by the triage automation for issue #614. A human reviewer should confirm the hook ordering and context-echo invariants before merging.*

https://claude.ai/code/session_015hHQqXRbn2jX9WTu564gjZ

---
_Generated by [Claude Code](https://claude.ai/code/session_015hHQqXRbn2jX9WTu564gjZ)_

### DIFF
--- a/src/adcp/decisioning/serve.py
+++ b/src/adcp/decisioning/serve.py
@@ -452,8 +452,9 @@ def serve(
             )
 
         Hook exceptions surface as ``INVALID_REQUEST`` on the wire.
-        The hook must return a new dict; mutating the input in-place is
-        a bug — the original is captured separately for context echo.
+        The hook receives a shallow copy of the wire args, so it may
+        mutate its argument freely or return a new dict — either style
+        is safe. Context echo always reflects the original wire input.
     """
     # Local import to avoid a circular at module-load time. Adopter
     # serves never run during foundation imports anyway.
--- a/src/adcp/decisioning/serve.py
+++ b/src/adcp/decisioning/serve.py
@@ -373,6 +373,7 @@ def serve(
     advertise_all: bool = False,
     mock_ad_server: Any | None = None,
     enable_debug_endpoints: bool = False,
+    pre_validation_hooks: dict[str, Any] | None = None,
     **serve_kwargs: Any,
 ) -> None:
     """One-call wrapper — build the handler and serve over MCP.
@@ -435,6 +436,24 @@ def serve(
         responses="strict")`` to enable schema-driven request/response
         validation against the bundled AdCP JSON schemas — sellers who
         want their server to enforce wire conformance turn it on here.
+    :param pre_validation_hooks: Optional dict mapping AdCP tool name to
+        a ``(tool_name, raw_args) -> raw_args`` callable. The hook runs
+        on the raw wire dict **before** schema + Pydantic validation —
+        use it to apply spec-mandated defaults for pre-v3 buyers that
+        omit required fields. Example::
+
+            serve(
+                router,
+                pre_validation_hooks={
+                    "get_products": lambda n, a: {
+                        **a, "buying_mode": a.get("buying_mode", "brief")
+                    },
+                },
+            )
+
+        Hook exceptions surface as ``INVALID_REQUEST`` on the wire.
+        The hook must return a new dict; mutating the input in-place is
+        a bug — the original is captured separately for context echo.
     """
     # Local import to avoid a circular at module-load time. Adopter
     # serves never run during foundation imports anyway.
@@ -504,6 +523,8 @@ def serve(
 
     server_name = name or type(platform).__name__
     debug_traffic_source = mock_ad_server.get_traffic if mock_ad_server is not None else None
+    if pre_validation_hooks is not None:
+        serve_kwargs["pre_validation_hooks"] = pre_validation_hooks
     _adcp_serve(
         handler,
         name=server_name,
--- a/src/adcp/server/a2a_server.py
+++ b/src/adcp/server/a2a_server.py
@@ -143,6 +143,7 @@ class ADCPAgentExecutor(AgentExecutor):
         message_parser: MessageParser | None = None,
         advertise_all: bool = False,
         validation: ValidationHookConfig | None = SERVER_DEFAULT_VALIDATION,
+        pre_validation_hooks: dict[str, Any] | None = None,
         test_controller_account_resolver: Any | None = None,
     ) -> None:
         self._handler = handler
@@ -169,7 +170,10 @@ class ADCPAgentExecutor(AgentExecutor):
             name = tool_def["name"]
             if name == "comply_test_controller" and test_controller is None:
                 continue
-            self._tool_callers[name] = create_tool_caller(handler, name, validation=validation)
+            hook = (pre_validation_hooks or {}).get(name)
+            self._tool_callers[name] = create_tool_caller(
+                handler, name, validation=validation, pre_validation_hook=hook
+            )
 
         if test_controller is not None:
             self._register_test_controller(test_controller)
@@ -758,6 +762,7 @@ def create_a2a_server(
     message_parser: MessageParser | None = None,
     advertise_all: bool = False,
     validation: ValidationHookConfig | None = SERVER_DEFAULT_VALIDATION,
+    pre_validation_hooks: dict[str, Any] | None = None,
     context_builder: Any | None = None,
     auth: BearerTokenAuth | None = None,
     public_url: str | None = None,
@@ -884,6 +889,7 @@ def create_a2a_server(
         message_parser=message_parser,
         advertise_all=advertise_all,
         validation=validation,
+        pre_validation_hooks=pre_validation_hooks,
         test_controller_account_resolver=test_controller_account_resolver,
     )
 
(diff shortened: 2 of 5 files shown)


bokelley:
## Adopter confirmation (filed #614)

Read the diff. **This kills our [core/middleware/spec_defaults.py](https://github.com/bokelley/salesagent/blob/main/core/middleware/spec_defaults.py) (273 LOC of JSON-RPC body bytes-rewriting).** Migration is straightforward — every default we apply maps to a single hook callable.

### Migration sketch on our side

\`\`\`python
serve(
    router,
    pre_validation_hooks={
        "get_products": _apply_get_products_defaults,
        "create_media_buy": _apply_auth_filled_defaults,
        "sync_creatives": _sync_creatives_defaults,  # asset-type backfill + format_id wrap
        "activate_signal": _apply_auth_filled_defaults,
        "sync_accounts": _apply_auth_filled_defaults,
    },
    ...
)
\`\`\`

Where each callable is a small pure function — same logic as today's bytes-level patchers, just operating on a Python dict instead of decoded JSON bytes. Net: ~50 LOC of helpers replacing 273 LOC of ASGI middleware.

### One design question

The hook signature is \`(tool_name, raw_args) -> raw_args\` — fine. But the PR docstring says **"The hook must return a new dict; mutating the input in-place is a bug — the original is captured separately for context echo."**

This is a footgun. \`{**args, "buying_mode": "brief"}\` is correct; \`args["buying_mode"] = "brief"; return args\` looks correct but breaks context echo silently.

Two options to mitigate:
1. **Defensive copy at the call site** before invoking the hook: \`hook(name, dict(params))\` — small allocation, eliminates the footgun entirely.
2. **Document the failure mode harder** with a runnable test/example showing what breaks.

Option 1 is the cleaner adopter contract — let people mutate freely, the framework guarantees context-echo safety. The cost is one dict shallow-copy per call which is negligible relative to schema validation right after.

### Cross-link with #623

[#623](https://github.com/adcontextprotocol/adcp-client-python/issues/623) ("Typed dispatcher rejects valid request when \`account\` is omitted") is the spec-correct fix for one of the specific cases this hook works around. They're complementary: this hook is the general-purpose adopter API for ALL spec defaults; #623 is the canonical fix for one specific buyer-omission case. Adopters running this hook for \`account\` placeholder injection can drop that hook entry once #623 lands. Worth a docstring note pointing readers there.

### Ready to adopt

Once this lands we'll bump and migrate in one PR. Will report back on whether the migration matches the predicted ~50 LOC delta.

bokelley:
Good catch on the footgun — Option 1 applied in 49feaef.

`create_tool_caller` now passes `dict(params)` to the hook instead of the live wire dict, so in-place mutation (`args["buying_mode"] = "brief"; return args`) is safe. The `raw_params` snapshot is taken before the copy is made, so context echo still reflects the original wire input regardless of what the hook does to its argument.

Docstrings in `mcp_tools.py`, `a2a_server.py`, and `decisioning/serve.py` updated to remove the "must return a new dict" restriction and describe the actual contract. Added `test_in_place_mutation_is_safe_for_context_echo` to lock the invariant — it exercises exactly the mutate-in-place-and-return pattern.

Also added the #623 cross-reference you suggested: adopters using a hook to inject an `account` placeholder can drop that entry once #623 lands.

Your migration sketch matches — `{**args, "buying_mode": "brief"}` and `args["buying_mode"] = "brief"; return args` are now equivalent.

---

Issue #615, opened 2026-05-10
feat(types): nested model_dump() resolution in response models


bokelley:
## Problem

Pydantic doesn't auto-call custom `model_dump()` on nested models — if a parent's children have `model_dump()` overrides, the parent's `super().model_dump()` produces the children via the default Pydantic serialization, not via the children's overrides. Adopters that need custom serialization on a child class (most commonly: tenant-internal fields excluded with `exclude=True`, or AdCP-spec-shaped overrides on extended types) end up writing this in every response model:

```python
class GetCreativesResponse(AdCPBaseModel):
    creatives: list[Creative]

    def model_dump(self, **kwargs):
        result = super().model_dump(**kwargs)
        if \"creatives\" in result and self.creatives:
            result[\"creatives\"] = [c.model_dump(**kwargs) for c in self.creatives]
        return result
```

We have **~59 such overrides** across [src/core/schemas/](https://github.com/prebid/salesagent/tree/main/src/core/schemas). Every override is mechanical: walk the children, call their `model_dump`. Adding a new response type means writing one again. Forgetting one is silent — the test fails on a serialization-difference, but only if the test exercises the nested path.

## Proposed SDK shape

Either of:

**A) `AdCPBaseModel` recursively dispatches custom `model_dump()` on children.** Walk `__pydantic_fields__`, detect when a field's annotated type has a custom `model_dump`, and call it directly instead of relying on Pydantic's default child serialization. Same for lists/dicts of models.

**B) Codegen flag** that emits nested-aware `model_dump()` on the generated response types — adopters extending those types via inheritance get the nested-walking for free.

(A) is broader; (B) is narrower but ships with no risk to non-adopters.

## Why this is the highest-LOC win on our open list

The other open framework gaps (spec defaults, agent-card URL, transport contextvar, test app builder) each kill 50-275 LOC of middleware. This one kills hundreds of LOC of mechanical override boilerplate spread across the schema package.

## Files

- Our overrides live across [src/core/schemas/_base.py](https://github.com/prebid/salesagent/blob/main/src/core/schemas/_base.py), product.py, creative.py, delivery.py, account.py.
- Pattern documented in our CLAUDE.md as "Critical Pattern #4: Pydantic Explicit Nested Serialization" — we'd love to delete the section.

bokelley:
## Triage

**Classification:** Feature request → resolved as documentation gap
[...]

PR #630, opened 2026-05-10
docs(types): document Field(exclude=True) and @model_serializer for nested wire isolation

bokelley:
Refs #615

Issue #615 asked for `AdCPBaseModel.model_dump()` to recursively dispatch `model_dump()` overrides on nested child instances. Expert analysis determined the root cause is a documentation gap: `Field(exclude=True)` and `@model_serializer` already solve both primary use cases — field-level exclusion and custom Python serialization logic — at every nesting depth. The 59 manual parent-dispatch overrides in downstream integrations can be deleted entirely using either pattern.

## What changed

**`docs/extending-types.md`** — three new sections added at the top of the guide:

- **Field-Level Exclusion with `Field(exclude=True)` — Recommended**: Shows how fields annotated with `Field(exclude=True)` are excluded by Pydantic's own Rust-backed serializer at every nesting depth, with no parent override required. Verified runnable example using `Creative`/`CreativePayload`.

- **Custom Serialization Logic with `@model_serializer`**: Shows `@model_serializer(mode='wrap')` for Python-level transformation logic. Includes an explicit `**Important:**` callout explaining the `serialize_as_any=True` requirement when a parent field is declared as the base type — common misconception that drove the pattern in #615.

- **Migrating from Manual `model_dump()` Dispatch Overrides**: Shows the old boilerplate pattern and both migration paths (field exclusion and custom logic).

Also updated:
- **Best Practices #1**: Elevated `Field(exclude=True)` over the previous "OK but fragile" call-site `model_dump(exclude={...})` recommendation.
- **Pre-existing type name fixes**: `CreateMediaBuySuccess` → `CreateMediaBuySuccessResponse` (correct export, 10+ sites), `WebhookPayload` → `McpWebhookPayload` (correct export). These were `ImportError`-level bugs in the pre-existing patterns.

**`src/adcp/types/base.py`** — 6-line comment on `AdCPBaseModel.model_dump()` explaining the Pydantic v2 Rust-serializer limitation and pointing to `docs/extending-types.md`.

## What was tested

```
ruff check src/adcp/types/base.py  # All checks passed
pytest tests/test_response_builder_subclass.py -v  # 5 passed, 2 warnings (pre-existing)
```

All three new doc examples verified runnable against pydantic 2.13.4 + actual SDK:
- `Field(exclude=True)` at nesting depth: internal fields absent ✓
- `@model_serializer` direct: serializer fires, `source_label` lowercased ✓
- `@model_serializer` nested without `serialize_as_any`: field absent ✓
- `@model_serializer` nested with `serialize_as_any=True`: field present, normalized ✓

## Pre-PR review

- **code-reviewer**: approved (pass 2) — `@model_serializer` claim now correct and verified; `Field(exclude=True)` claim correct at all nesting depths; no remaining blockers.
- **docs-expert**: approved (pass 2) — all 5 original blockers resolved: correct import path (`adcp.types.base`), valid `Creative` constructor, `source_label` replaces non-existent `render_url`, `MyCreativePayload` replaces non-existent `GetCreativesResponse`, `CreateMediaBuySuccessResponse` in Best Practices #1.

**Nits (not fixed, noted for follow-up):**
- `Summary` section recommends `ConfigDict(extra='allow')` without a caveat that this admits unknown wire fields; a one-line note distinguishing "use only when intentionally accepting unknown fields" would help.
- `model_dump_json()` on `AdCPBaseModel` has the same Pydantic v2 nesting behavior but no matching comment.
- `docs/handler-authoring.md:697` could add a forward-reference to the new `Field(exclude=True)` section.

---

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01P7MQW9tW7z4rYm13zghrVC

---
_Generated by [Claude Code](https://claude.ai/code/session_01P7MQW9tW7z4rYm13zghrVC)_

### DIFF
--- a/docs/extending-types.md
+++ b/docs/extending-types.md
@@ -8,8 +8,9 @@ This guide shows how to extend ADCP types safely while maintaining protocol comp
 > model calls `model_dump()`, Pydantic serializes nested child instances using its own compiled
 > pipeline — it does **not** call Python-level `model_dump()` overrides on child objects. This
 > means that if you override `model_dump()` on a child class, that override will not fire when
-> the child is serialized as part of a parent. Use `Field(exclude=True)` or `@model_serializer`
-> instead (both covered below).
+> the child is serialized as part of a parent. Use `Field(exclude=True)` for field-level
+> exclusion (works automatically at every nesting depth) or `@model_serializer` with
+> `serialize_as_any=True` for custom Python logic (covered below).
 
 ## Field-Level Exclusion with `Field(exclude=True)` — Recommended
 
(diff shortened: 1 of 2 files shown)


bokelley:
## Adopter pushback — closing #615 with docs-only is incomplete

The doc improvements are genuinely good (especially elevating \`Field(exclude=True)\` over call-site \`exclude={}\` plumbing — that part is the right pattern and we're already using it: 58 \`exclude=True\` declarations in our schema package). **But the docs themselves admit the actual gap, and the resolution is documenting the workaround instead of fixing it.**

### The contradiction

The PR's own docstring contains:

> When a parent field is annotated as the base type (\`creatives: list[Creative]\`), Pydantic's Rust serializer uses the declared type's schema and the subclass \`@model_serializer\` does **not** fire automatically. You must pass \`serialize_as_any=True\` to the parent's \`model_dump()\` call to apply subclass serializers from a parent.

This is **exactly the original problem** #615 reported. The issue wasn't "we don't know how to write exclusions" — we already use \`Field(exclude=True)\` extensively (58 sites). The issue was that **Pydantic's nested-dispatch behaviour requires every \`model_dump()\` caller in the stack to remember \`serialize_as_any=True\`**, which is precisely the kind of error-prone boilerplate adopters wanted eliminated.

### What our 14 surviving overrides actually look like

Audited our schema package. The pattern is:

\`\`\`python
@model_serializer(mode="wrap")
def _serialize_model(self, serializer, info):
    data = serializer(self)
    # Walk children to re-apply their serializers because Pydantic won't
    for field_name, _ in self.__class__.model_fields.items():
        field_value = getattr(self, field_name, None)
        if isinstance(field_value, list) and field_value and isinstance(field_value[0], BaseModel):
            data[field_name] = [item.model_dump(mode=info.mode) for item in field_value]
        elif isinstance(field_value, BaseModel):
            data[field_name] = field_value.model_dump(mode=info.mode)
    return data
\`\`\`

This boilerplate exists in ~14 of our parent models. **The PR's recommended migration is to delete these and instead pass \`serialize_as_any=True\` at every call site.** That's a worse contract than the boilerplate — at least the boilerplate is enforced once at the model level. \`serialize_as_any=True\` at every call site is unenforceable; missing one is silent.

### Real fix: default \`serialize_as_any=True\` in \`AdCPBaseModel.model_dump()\`

\`AdCPBaseModel.model_dump\` already injects one default:

\`\`\`python
def model_dump(self, **kwargs: Any) -> dict[str, Any]:
    if "exclude_none" not in kwargs:
        kwargs["exclude_none"] = True
    return super().model_dump(**kwargs)
\`\`\`

Adding one more line:

\`\`\`python
    if "serialize_as_any" not in kwargs:
        kwargs["serialize_as_any"] = True
\`\`\`

…makes subclass \`@model_serializer\` and subclass-only \`Field(exclude=True)\` fire transparently for every \`AdCPBaseModel\` descendant. **That's the actual fix for #615** — the docstring caveat goes away, the ~14 manual parent overrides can be deleted, and adopters never have to remember the kwarg.

Side-effect risk: \`serialize_as_any=True\` would expose subclass-only fields that the parent's declared type doesn't know about. **This is desirable** for adopters following the documented pattern (subclass adds field with \`exclude=True\` for internal, or without for genuinely-extended wire shapes). For adopters who don't follow the pattern, the failure mode is "fields that should be excluded leak" — surfaced loudly at the first wire test, not silently like the current \`serialize_as_any=False\` default.

### Recommendation

1. **Merge this PR for the doc improvements** — \`Field(exclude=True)\` recommendation is right.
2. **Don't close #615.** The doc workaround isn't a fix; it's a workaround for a one-line SDK change.
3. **File a follow-up (or extend this PR) to default \`serialize_as_any=True\` on \`AdCPBaseModel.model_dump()\`.** The doc caveat is the strongest possible argument for the change — every adopter reading those docs has to internalise an arbitrary kwarg invariant that the SDK could just default.

If you want, I can spike the one-line change against \`src/adcp/types/base.py\` and run the test suite — happy to confirm whether anything actually depends on the current \`serialize_as_any=False\` default. My read is nothing should, but worth verifying.

### Nits in the diff

- \`from adcp import McpMcpWebhookPayload\` (line ~287 of the doc diff) — typo, double "Mcp".
- \`InternalMcpWebhookPayload.model_validate(payload)\` is referenced one line before \`InternalWebhookPayload\` is defined — find/replace miss when renaming the class.
- The Summary section's \`ConfigDict(extra='allow')\` recommendation lacks a security caveat (this is the nit your own pre-PR review flagged).

bokelley:
Re-ran code-reviewer + security-reviewer on the proposed one-liner for #615. The thesis is correct — but two security blockers surface with a **global** default injection.

**Ship-blockers:**

1. **`dispatch.py:1701` — task registry persistence.** `result.model_dump()` (bare, no args) runs on the `handoff_to_task` return path. With a global `serialize_as_any=True`, adopter subclass fields not marked `Field(exclude=True)` — `approval_token`, `internal_order_id`, upstream credentials — land in the idempotency registry and are returned verbatim via `tasks/get`. `strip_credentials_from_wire_result` covers only two spec-known fields; adopter-specific names pass through.

2. **`dispatch.py:538` — request context echo.** `params.model_dump(mode="json", exclude_none=False)` is stored as `request_context` on `TaskRecord` and echo'd in `tasks/get` responses. Subclass request fields bypass `_validate_ctx_metadata_credentials` entirely.

**Recommended narrower fix:** Apply `serialize_as_any=True` at the specific call sites in `adcp.server.responses._serialize()` (the tested wire-response path), not globally via `AdCPBaseModel.model_dump()`. This gets the subclass-passthrough guarantee where it's tested and wanted, while leaving the task-registry and request-echo paths unexpanded.

**Also flagged (non-blockers):** `model_dump_json()` would have asymmetric behavior; docs in `extending-types.md` would describe the old behavior post-fix; the base.py comment needs updating to reflect the new default.

**Status: ready-for-human.** @bokelley — preferred path: targeted call-site injection in `_serialize()`, or global default with explicit `serialize_as_any=False` guards at the two dispatch call sites?

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01FtfxrtWLU5Pg7du7hhuj2Y

---

PR #633, opened 2026-05-10
fix(types): make account optional in CreateMediaBuyRequest and SyncCreativesRequest

### LINKED ISSUE #623: Typed dispatcher rejects valid request when `account` is omitted on auth-resolved tools

bokelley:
Closes #623

## Summary

The typed dispatcher was rejecting valid `create_media_buy` and `sync_creatives` requests when the buyer omitted `account`. Per AdCP 3.0.6+, `account` is optional when the seller uses implicit or derived resolution mode — account identity is resolved from the verified auth chain instead of the wire field.

Root cause: `CreateMediaBuyRequest.account` and `SyncCreativesRequest.account` were defined as required Pydantic fields (no default). Pydantic rejected the request with `INVALID_REQUEST` before the handler could call `AccountStore.resolve`, which already accepts `ref=None` and correctly derives account identity for implicit/derived mode.

Changes:
- `src/adcp/types/generated_poc/media_buy/create_media_buy_request.py`: `account: AccountReference` → `account: AccountReference | None = None`
- `src/adcp/types/generated_poc/creative/sync_creatives_request.py`: same
- `schemas/cache/{media-buy,creative}/` + `bundled/` variants: removed `account` from `required[]` so `generate_types.py` regen reproduces the optional field on next stable-tag run
- `tests/test_issue_623_implicit_account_optional.py`: 7 regression tests (model-level + wire dispatch for derived mode + explicit-mode fail-closed)
- `tests/test_mcp_schema_drift.py`: updated spot-check to assert `account` is NOT in `create_media_buy.required`

`GetProductsRequest.account` was already `AccountReference | None = None` — this change aligns the other two request types with that existing pattern.

`ExplicitAccounts.resolve` still raises `ACCOUNT_NOT_FOUND` when `ref=None`, preserving fail-closed behavior for explicit-mode adopters without any handler changes.

## What was tested

- `ruff check src/` — clean
- `mypy` on changed files — no new errors (pre-existing import-not-found for httpx/a2a stubs unrelated)
- `pytest tests/test_issue_623_implicit_account_optional.py` — 7/7 passed
- `pytest tests/test_mcp_schema_drift.py` — 19/19 passed
- `pytest tests/ --ignore=tests/conformance/signing/` — 3773 passed, 24 skipped, 1 xfailed (pre-existing TLS network test unrelated to this change)

## Pre-PR review

- code-reviewer: approved after 2-round review — both blockers fixed (schema cache `required[]` + model_construct→model_validate tests); no new blockers
- ad-tech-protocol-expert: sound — change aligns with `GetProductsRequest` precedent; `AccountStore.resolve(ref=None)` contract already correct for implicit/derived; `ExplicitAccounts` fail-closed path preserved

**Nits surfaced (not fixed):**
- `lambda account_id: None` in `_ExplicitModePlatform` fixture could be `lambda _: None`
- Local `from adcp.decisioning.types import AdcpError` import inside test function could move to module-level imports
- `test_mcp_schema_drift.py` spot-check has no parallel `sync_creatives` assertion (covered by model-level test instead)
- `idempotency_key` description still reads "unique per (seller, request) pair" — with implicit mode the scope is actually `(store_qualname, account.id)` where `account.id` is synthesized by the store (pre-existing description, out of this PR's scope)

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01JX6bQ9FK2YxLefGnUSkg6x

---
_Generated by [Claude Code](https://claude.ai/code/session_01JX6bQ9FK2YxLefGnUSkg6x)_

### DIFF
--- a/schemas/cache/bundled/creative/sync-creatives-request.json
+++ b/schemas/cache/bundled/creative/sync-creatives-request.json
@@ -4916,7 +4916,6 @@
   },
   "required": [
     "idempotency_key",
-    "account",
     "creatives"
   ],
   "additionalProperties": true,
--- a/schemas/cache/bundled/media-buy/create-media-buy-request.json
+++ b/schemas/cache/bundled/media-buy/create-media-buy-request.json
@@ -7445,7 +7445,6 @@
   },
   "required": [
     "idempotency_key",
-    "account",
     "brand",
     "start_time",
     "end_time"
--- a/schemas/cache/creative/sync-creatives-request.json
+++ b/schemas/cache/creative/sync-creatives-request.json
@@ -108,7 +108,6 @@
   },
   "required": [
     "idempotency_key",
-    "account",
     "creatives"
   ],
   "additionalProperties": true,
--- a/schemas/cache/media-buy/create-media-buy-request.json
+++ b/schemas/cache/media-buy/create-media-buy-request.json
@@ -206,7 +206,6 @@
   },
   "required": [
     "idempotency_key",
-    "account",
     "brand",
     "start_time",
     "end_time"
--- a/src/adcp/types/generated_poc/creative/sync_creatives_request.py
+++ b/src/adcp/types/generated_poc/creative/sync_creatives_request.py
@@ -53,8 +53,11 @@ class SyncCreativesRequest(AdCPBaseModel):
         ),
     ] = None
     account: Annotated[
-        account_ref.AccountReference, Field(description='Account that owns these creatives.')
-    ]
+        account_ref.AccountReference | None,
+        Field(
+            description='Account that owns these creatives. May be omitted when the seller uses implicit or derived resolution mode — account identity is then resolved from the verified auth chain.'
+        ),
+    ] = None
     creatives: Annotated[
         list[creative_asset.CreativeAsset],
         Field(
--- a/src/adcp/types/generated_poc/media_buy/create_media_buy_request.py
+++ b/src/adcp/types/generated_poc/media_buy/create_media_buy_request.py
@@ -152,11 +152,11 @@ class CreateMediaBuyRequest(AdCPBaseModel):
         ),
     ] = None
     account: Annotated[
-        account_ref.AccountReference,
+        account_ref.AccountReference | None,
         Field(
-            description='Account to bill for this media buy. Pass a natural key (brand, operator, optional sandbox) or a seller-assigned account_id from list_accounts.'
+            description='Account to bill for this media buy. Pass a natural key (brand, operator, optional sandbox) or a seller-assigned account_id from list_accounts. May be omitted when the seller uses implicit or derived resolution mode — account identity is then resolved from the verified auth chain.'
         ),
-    ]
+    ] = None
     proposal_id: Annotated[
         str | None,
         Field(
--- a/tests/test_issue_623_implicit_account_optional.py
+++ b/tests/test_issue_623_implicit_account_optional.py
@@ -1,0 +1,174 @@
+"""Regression tests for issue #623.
+
+The typed dispatcher was rejecting valid requests on ``create_media_buy``
+and ``sync_creatives`` when the buyer omitted ``account``. Per the AdCP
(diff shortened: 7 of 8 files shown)


bokelley:
## Closing — the premise of this PR is wrong

After re-deriving the resolution-mode semantics with @bokelley:

- **explicit mode** — seller owns the account catalog; supports both `account.account_id` (seller-assigned) and natural-key (brand+operator → resolve).
- **derived mode** — account catalog comes from the decisioning platform (GAM, etc.); buyer addresses by the **platform's `account_id`** returned from `list_accounts`. Natural keys don't apply because the seller doesn't own the mapping — the platform does.

Either way, the buyer **must** identify which account they're transacting against. There's no resolution mode where the dispatcher should accept `create_media_buy` / `sync_creatives` with `account` omitted — there's no clean way to tell "which advertiser's budget to debit" without an explicit ref.

This PR makes `account: AccountReference | None = None`, which doesn't model a real spec mode. Closing as won't-merge. The salesagent `auth-chain` placeholder middleware that motivated #623 is itself hiding a buyer-side protocol violation (buyers should be passing a real `account_id` they got from `list_accounts`), and the right fix lives there, not in the SDK types.

See updated discussion on #623.

bokelley:
Acknowledged. The protocol reasoning is sound — regardless of resolution mode (explicit or derived), the buyer must supply an `AccountReference` so the dispatcher knows which account to debit. Making the field `None`-able doesn't correspond to any real spec mode and would silently accept malformed requests.

Closing this draft is the right call. I can't close PRs directly, so a repo member will need to hit the close button. The source issue #623 should be updated to reflect that the fix belongs in the salesagent auth-chain middleware (buyer should pass the `account_id` returned from `list_accounts`), not in the SDK field optionality.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_01EnjvAYLJHWtayiyP9My91X_

---

PR #640, opened 2026-05-10
feat(types): widen extension-point list[X] to Sequence[X] (#624)

### LINKED ISSUE #624: feat(types): adopter inheritance pattern triggers ~50 type: ignore[assignment]
## Problem

The recommended adopter pattern — extend a library type and override a nested field with a more-specific element type — requires `# type: ignore[assignment]` to satisfy mypy under `strict = true`. We carry roughly 50 of these in our codebase across the schema package.

The mypy Pydantic plugin treats field annotations as invariant under inheritance, even when the override is logically covariant (every `AffectedPackage` IS-A `LibraryAffectedPackage`).

## Reproducer

```python
from adcp.types import UpdateMediaBuyResponse as LibraryUpdateMediaBuyResponse
from adcp.types import AffectedPackage as LibraryAffectedPackage
from pydantic import Field

class AffectedPackage(LibraryAffectedPackage):
    # Internal-only fields excluded on the wire
    changes_applied: dict | None = Field(default=None, exclude=True)
    buyer_package_ref: str | None = Field(default=None, exclude=True)

class UpdateMediaBuyResponse(LibraryUpdateMediaBuyResponse):
    # Override to use the extended element type — same shape on the wire,
    # extra internal fields excluded via exclude=True. Pydantic accepts it
    # at runtime; mypy rejects it.
    affected_packages: list[AffectedPackage] | None = None  # type: ignore[assignment]
```

`mypy --strict` complains about `affected_packages` being incompatible with the parent's `list[LibraryAffectedPackage]`. There's nothing wrong at runtime — every `AffectedPackage` is a `LibraryAffectedPackage`. The issue is that mypy sees `list[X]` as invariant in `X`.

## Where this hits us

We carry ~50 instances of `# type: ignore[assignment]` in [src/core/schemas/](https://github.com/prebid/salesagent/tree/main/src/core/schemas) for exactly this pattern. Sample:

- [_base.py:360](https://github.com/prebid/salesagent/blob/main/src/core/schemas/_base.py#L360) — `affected_packages: list[AffectedPackage] | None`
- [_base.py:875-878](https://github.com/prebid/salesagent/blob/main/src/core/schemas/_base.py#L875) — geo exclusion lists
- [delivery.py:251, 440](https://github.com/prebid/salesagent/blob/main/src/core/schemas/delivery.py#L251)

This is the recommended pattern (Critical Pattern #1 in our CLAUDE.md). It's not a code smell on our side — we're following the docs. The strictness ergonomics for adopters are the issue.

## Proposed SDK shape

A few possible directions, in increasing order of impact:

**A) Use `Sequence[X]` instead of `list[X]`** in library response types where adopters are expected to override the element. `Sequence` is covariant in its element type, so `Sequence[AffectedPackage]` is a valid subtype of `Sequence[LibraryAffectedPackage]`.

Drawback: callers who want `.append()` need a mutable type. May not work for in-place mutation paths.

**B) Generic library types parameterised on extension points.** The library generates `UpdateMediaBuyResponse[TAffectedPackage = AffectedPackage]`, and adopters subclass with `class UpdateMediaBuyResponse(LibraryUpdateMediaBuyResponse[AffectedPackage]): ...`. mypy is happy; no `type: ignore` needed.

Drawback: more codegen complexity; new public API.

**C) Codegen extension hooks.** datamodel-codegen emits `# extension: list-element-type` comments on field annotations that adopters typically override; a separate codegen step generates an `Extended<Type>` skeleton that adopters can subclass with the field overridden cleanly.

Drawback: most invasive.

## Why this matters beyond us

Any Python adopter who extends library types under strict mypy hits this. We absorb 50 `type: ignore` lines because we're the most advanced adopter; subsequent adopters either (a) carry the same lines, (b) silence the errors with bare `type: ignore`, or (c) abandon the inheritance pattern in favour of duplication — which the docs explicitly warn against.

This is a strictness-ergonomics gap, not a runtime bug. Companion to #615 (nested `model_dump()` overrides), which is the same family of friction.

## Files

- Our overrides: [src/core/schemas/_base.py](https://github.com/prebid/salesagent/blob/main/src/core/schemas/_base.py), [delivery.py](https://github.com/prebid/salesagent/blob/main/src/core/schemas/delivery.py), [creative.py](https://github.com/prebid/salesagent/blob/main/src/core/schemas/creative.py)
- Pattern documented: CLAUDE.md "Critical Pattern #1: AdCP Schema: Extend Library Schemas"

bokelley:
Closes #624.

## Summary

Adopters following the documented Critical Pattern #1 (subclass a library response type, override a parent's \`list[X]\` field with \`list[ChildX]\` carrying internal-only fields) carry \`# type: ignore[assignment]\` on every override under mypy --strict. \`list[T]\` is invariant in T; \`Sequence[T]\` is covariant, so a \`Sequence[Parent]\` parent permits \`list[Child]\` override cleanly.

This PR adds a post-codegen processor that rewrites \`list[X]\` → \`Sequence[X]\` for fields on a small allowlist of documented extension points.

## What's in the diff

| File | Change |
|---|---|
| \`scripts/post_generate_fixes.py\` | New \`widen_extension_point_lists_to_sequence()\` + helpers, wired into \`main()\`. Allowlist keyed on (file, class, field). |
| \`src/adcp/types/generated_poc/media_buy/update_media_buy_response.py\` | \`UpdateMediaBuyResponse1.affected_packages\` widened. \`Sequence\` import added. |
| \`src/adcp/types/generated_poc/bundled/media_buy/update_media_buy_response.py\` | \`UpdateMediaBuyResponse3.affected_packages\` widened. \`Sequence\` import added. |
| \`tests/type_checks/extend_response_with_sequence.py\` | Regression gate — subclasses \`UpdateMediaBuyResponse1\` with overridden \`list[_InternalPackage]\` and asserts mypy --strict accepts it. |

## Allowlist status

**Confirmed (1):**
- \`UpdateMediaBuyResponse.affected_packages\` — matches salesagent \`_base.py:360\`. Two emitted variants both rewritten.

**Placeholders (9 TODOs):**
- \`_base.py:875-878\` (geo exclusion lists)
- \`delivery.py:251\`, \`delivery.py:440\`
- \`creative.py\` list_creatives asset list
- \`_base.py\` get_products products list
- \`_base.py\` sync_creatives result list
- \`_base.py\` list_accounts accounts list
- \`_base.py\` get_signals signals list
- \`_base.py\` list_creative_formats formats list

**Blocker on filling these in:** I need salesagent's actual \`# type: ignore[assignment]\` line list to map them accurately. Once the list is dropped, expanding the allowlist is a 5-line change per entry.

## Validation

- 2-file mypy --strict spike confirmed Pydantic plugin accepts \`Sequence[Parent]\` + \`list[Child]\` override (see [issue comment](https://github.com/adcontextprotocol/adcp-client-python/issues/624#issuecomment-4415735864)).
- Regression test \`tests/type_checks/extend_response_with_sequence.py\` runs under \`mypy --strict\` and proves \`.append()\` ergonomics survive on the child class.
- Local \`mypy src/adcp/\` clean (784 source files, no issues).
- Existing tests untouched.

## Breaking change

Callers passing \`affected_packages\` into a function typed \`def f(x: list[Package])\` will see a mypy error and need to migrate to \`Sequence[Package]\`. Runtime behavior is unchanged — the change is annotation-only.

## Open questions for review

1. Allowlist scope: should I expand the placeholders now with best-guess field paths, or wait for the actual salesagent list to avoid rework?
2. \`_ergonomic.py\`: BeforeValidator updates probably aren't needed (Sequence accepts list at construction), but worth a sanity check.
3. Naming: \`widen_extension_point_lists_to_sequence\` is verbose — open to a shorter name.

### DIFF
--- a/scripts/post_generate_fixes.py
+++ b/scripts/post_generate_fixes.py
@@ -875,7 +875,14 @@ def widen_extension_point_lists_to_sequence():
 
     # Track total rewrites per (class, field) — a pair with zero hits is
     # a stale allowlist entry and surfaces as a WARN.
+    # Track per-pair state across all files:
+    #   rewrites: how many list[X] sites were rewritten this run
+    #   already_widened: how many sites are already in Sequence[X] form
+    # A pair with rewrites == 0 AND already_widened == 0 is genuinely stale
+    # (field renamed/removed) and warrants a WARN. A pair with already_widened
+    # > 0 is silent — that's the steady-state idempotent run.
     rewrites_per_pair: dict[tuple[str, str], int] = {pair: 0 for pair in _SEQUENCE_EXTENSION_POINTS}
+    already_per_pair: dict[tuple[str, str], int] = {pair: 0 for pair in _SEQUENCE_EXTENSION_POINTS}
     files_touched = 0
     total_widened = 0
 
@@ -893,6 +900,8 @@ def widen_extension_point_lists_to_sequence():
                 content = new_content
                 widened_in_file += 1
                 rewrites_per_pair[(class_name, field_name)] += 1
+            elif _field_already_widened(content, class_name, field_name):
+                already_per_pair[(class_name, field_name)] += 1
 
         if widened_in_file == 0:
             continue
@@ -903,11 +912,15 @@ def widen_extension_point_lists_to_sequence():
         total_widened += widened_in_file
         print(f"  ✓ {file_path.relative_to(OUTPUT_DIR)}: widened {widened_in_file} field(s)")
 
-    stale = [pair for pair, n in rewrites_per_pair.items() if n == 0]
+    stale = [
+        pair
+        for pair in _SEQUENCE_EXTENSION_POINTS
+        if rewrites_per_pair[pair] == 0 and already_per_pair[pair] == 0
+    ]
     for class_name, field_name in stale:
         print(
-            f"  WARN: {class_name}.{field_name} — no list[X] occurrences found "
-            "(field renamed, removed, or already widened?)"
+            f"  WARN: {class_name}.{field_name} — neither list[X] nor Sequence[X] "
+            "found in any generated file (field renamed or removed?)"
         )
 
     if total_widened == 0:
@@ -923,11 +936,12 @@ def _widen_field_annotation(content: str, class_name: str, field_name: str) -> t
     """Rewrite ``list[X]`` → ``Sequence[X]`` in one field's annotation.
 
     Locates ``class {class_name}(...):`` then walks forward to the first
-    ``    {field_name}:`` line at class-body indentation. Within the
-    AnnAssign's annotation block (which may span multiple lines for
-    ``Annotated[..., Field(...)]``), replaces the first ``list[`` with
-    ``Sequence[``. Idempotent: a second pass over Sequence-form content
-    is a no-op.
+    ``    {field_name}:`` line at class-body indentation, **bounded to the
+    target class** so a same-named field on a later class in the same
+    file cannot mis-match. Within the AnnAssign's annotation block (which
+    may span multiple lines for ``Annotated[..., Field(...)]``), replaces
+    the first ``list[`` with ``Sequence[``. Idempotent — a second pass
+    over already-widened content is a no-op.
     """
     # Anchor on the class definition.
     class_pattern = re.compile(rf"^class {re.escape(class_name)}\b", re.MULTILINE)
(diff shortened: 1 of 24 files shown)


bokelley:
## Allowlist expanded to all 15 Category A entries

Refactored the rewriter from \`(file, class, field)\` to \`(class, field)\` and made it walk every generated \`.py\` file. \`datamodel-codegen\` emits bundled response files that each inline copies of subordinate types (\`Placement\`, \`TargetingOverlay\`, etc.) — every emission gets rewritten so all paths stay consistent.

**47 fields widened across 21 generated files.**

| Library type | Field |
|---|---|
| \`UpdateMediaBuyResponse\` (1, 3) | \`affected_packages\` |
| \`GetMediaBuyDeliveryResponse\` | \`media_buy_deliveries\` |
| \`GetCreativeDeliveryResponse\` | \`creatives\` |
| \`Signal\` | \`deployments\` |
| \`GetSignalsResponse\` | \`signals\` |
| \`GetMediaBuysResponse\` | \`media_buys\` |
| \`ListCreativesResponse\` | \`creatives\` |
| \`PackageRequest\` | \`creatives\` |
| \`CreateMediaBuyRequest\` | \`packages\` |
| \`UpdateMediaBuyRequest\` | \`packages\` |
| \`Placement\` | \`format_ids\` |
| \`TargetingOverlay\` | \`geo_countries_exclude\`, \`geo_regions_exclude\`, \`geo_metros_exclude\`, \`geo_postal_areas_exclude\` |

Local checks: \`mypy src/adcp/\` clean (784 source files), \`mypy --strict tests/type_checks/\` clean (9 source files), test sanity sweep on \`test_decisioning_dispatch\` + \`test_pre_validation_hooks\` 63 passed.

The walker also emits a \`WARN\` for any \`(class, field)\` pair that doesn't match anywhere in the tree, so future allowlist drift (renamed/removed fields) surfaces fast on regen.

CI rolling now.

bokelley:
Noted — scope expansion looks solid. The `(class, field)` keyed walker covering all generated files is the right call given `datamodel-codegen`'s bundled-emission pattern, and the drift-warning on unmatched allowlist pairs is a good safety net for future regens. Standing by for CI green; happy to re-run expert review on the expanded diff once it lands.

---

bokelley:
**Fold candidate from #642:** Issue #642 (codegen type-identity mismatch — `Creative` re-emitted per response file, so `adcp.types.Creative` and the field-local `Creative` are type-distinct under mypy) was just filed as a companion to #624. The proposed fix is a post-processor step in `post_generate_fixes.py` that replaces local re-emissions with cross-file imports from the canonical module (Approach B from the issue). File scope overlaps with this PR's existing `widen_extension_point_lists_to_sequence()` addition. If you're still iterating here, folding #642's allowlist rewrite in before merge would close both halves of the adopter-override friction in one PR — no `type: ignore[assignment]` on type-identity grounds, no `type: ignore[assignment]` on variance grounds.

---

bokelley:
Investigated the fold. The two `Creative` classes aren't nominally identical — they're genuinely different schemas:

- `get_creative_delivery_response.Creative` (what `adcp.types.Creative` resolves to): delivery-metrics type — `creative_id`, `media_buy_id`, `totals`, `variant_count`, `variants`
- `list_creatives_response.Creative`: full creative-asset record — `name`, `status`, `assets`, `tags`, `variables`, `assignments`, etc.

Approach B as written (replace local re-emission with an import from the canonical module) would wire the delivery-metrics `Creative` into `ListCreativesResponse.creatives`, which is structurally wrong. The real problem in #642 is that codegen assigned the name `Creative` to two distinct types, and `_generated.py` happens to export the delivery one as the public alias — not a simple identity mismatch fixable with a cross-file import.

Recommendation: proceed with #640 as-is (Sequence widening still eliminates the variance half of the friction) and leave #642 open as a separate track requiring either a codegen rename (`CreativeDelivery` vs `Creative`) or a `_generated.py` export fix to surface the right canonical type.

---

bokelley:
Understood — the investigation confirms these are genuinely distinct schemas, not an identity mismatch, so the fold would have wired the wrong type. #640 ships the Sequence widening as intended; #642 stays open as a separate codegen-rename track (`CreativeDelivery` vs `Creative` or a `_generated.py` export fix). No further action needed here.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_01FGyjS99jvAz6F58M9qpH35_

---

PR #679, opened 2026-05-11
fix(compat): extract hostname from brand_manifest URLs with paths

### LINKED ISSUE #677: v2.5 brand_manifest adapter preserves URL path, breaks brand.domain regex
## Repro

Send a v2.5 ``get_products`` request with a ``brand_manifest`` URL that includes a path:

```python
client.call_tool("get_products", {"brand_manifest": "https://acme.com/.well-known/brand.json", "brief": "..."})
```

## Observed

```
INVALID_REQUEST[brand.domain]: get_products failed:
Invalid value for field 'brand.domain':
String should match pattern '^[a-z0-9]([a-z0-9-]*[a-z0-9])?(\.[a-z0-9]([a-z0-9-]*[a-z0-9])?)*$'
```

## Root cause

In `adcp/compat/legacy/v2_5/get_products.py:130-133`:

```python
# brand_manifest (v2.5 URL string) → brand.domain (v3 BrandReference)
brand_manifest = out.pop("brand_manifest", None)
if isinstance(brand_manifest, str) and brand_manifest and "brand" not in out:
    out["brand"] = {"domain": strip_url_scheme(brand_manifest)}
```

`strip_url_scheme` (in `adcp/compat/legacy/v2_5/_url.py`) only strips the scheme prefix and trailing slash. For `https://acme.com/.well-known/brand.json` it returns `acme.com/.well-known/brand.json` — which has slashes and dots in positions that fail `BrandReference.domain`'s regex.

The SDK's own `_normalize_brand_manifest` in `adcp/server/translate.py:485` (the utility intended for adopters to call manually) gets this right — it uses `urlparse(manifest).hostname`. Same concept, different impl, only one of the two extracts the hostname correctly.

## Spec context

`brand_manifest` per the v2.5 spec is documented as `"https://example.com/brand.json"` — i.e. a URL to a JSON file, which by construction has a path. So the legitimate input shape is exactly the one this adapter breaks.

## Impact

Surfaced via shape-based legacy detection (#673) in 5.2.0 — any v2.5 buyer (untagged or otherwise) sending a `brand_manifest` URL with a path gets a confusing `brand.domain` regex error. Trapped us on a CI run where the test had been passing under 5.1.0 (where `brand_manifest` got silently stripped as an unknown field, no translation attempted).

## Suggested fix

```python
from urllib.parse import urlparse

if isinstance(brand_manifest, str) and brand_manifest and "brand" not in out:
    parsed = urlparse(brand_manifest)
    domain = parsed.hostname or strip_url_scheme(brand_manifest)
    out["brand"] = {"domain": domain}
```

Same shape as `_normalize_brand_manifest` in `translate.py`. The `strip_url_scheme` fallback covers the case where the buyer sent a bare domain without `://` (no scheme → `hostname` is `None`).

Also applies to `adapt_request` in `create_media_buy.py` and `update_media_buy.py` if they take the same path (haven't checked).

## Environment

- `adcp==5.2.0`
- v2.5 wire-version buyer (or shape-detected via `is_legacy_shape`)

bokelley:
Closes #677

## Summary

`strip_url_scheme` only strips the scheme prefix and trailing slash — for a v2.5 `brand_manifest` URL like `"https://acme.com/.well-known/brand.json"` it returned `"acme.com/.well-known/brand.json"`, which fails `BrandReference.domain`'s hostname-only regex and surfaced as a confusing `INVALID_REQUEST[brand.domain]` error. The v2.5 spec documents `brand_manifest` as a URL to a JSON file, so a path is the *expected* input shape.

This PR adds `extract_brand_domain()` to `_url.py` using `urlparse().hostname`, which correctly strips both path and port. A bare-domain fallback via `strip_url_scheme` handles callers that pass a schemeless string. Both adapter call sites — `get_products.adapt_request` and `_media_buy_helpers.adapt_brand_manifest_to_brand` (used by `create_media_buy`) — are updated. `update_media_buy` is not affected (it carries no `brand_manifest`).

## What was tested

- `pytest tests/test_legacy_adapter_v2_5_get_products.py tests/test_legacy_adapter_v2_5_media_buy.py tests/test_dispatcher_shape_detection.py -v` — **54 passed**
- `ruff check src/adcp/compat/` — no issues
- `mypy src/adcp/compat/` — no issues in 12 source files

- Schemeless URL with path (e.g. `"acme.com/.well-known/brand.json"`) falls through to `strip_url_scheme` and produces an invalid domain — pre-existing gap, not introduced here; v2.5 spec defines `brand_manifest` as a full URL so this input shape is malformed
- `strip_url_scheme` retains a public name despite now being an internal fallback inside `extract_brand_domain`; consider renaming to `_strip_url_scheme` in a follow-up
- IPv6 literal hostnames are extracted without brackets by `urlparse` and will still fail `BrandReference.domain`'s regex; documented in `extract_brand_domain`'s docstring, callers must validate if IPv6 is possible (v2.5 spec makes this unrealistic for `brand_manifest`)

## Pre-PR review

- code-reviewer: approved — no blockers; schemeless-with-path fallback gap is pre-existing and documented
- dx-expert: approved — fix is correct at both call sites; nits on `strip_url_scheme` naming and IPv6 noted above

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01YX14H75zYWQ1BYuopyUL34

---
_Generated by [Claude Code](https://claude.ai/code/session_01YX14H75zYWQ1BYuopyUL34)_

### DIFF
--- a/src/adcp/compat/legacy/v2_5/_media_buy_helpers.py
+++ b/src/adcp/compat/legacy/v2_5/_media_buy_helpers.py
@@ -14,17 +14,22 @@ from __future__ import annotations
 
 from typing import Any
 
-from adcp.compat.legacy.v2_5._url import strip_url_scheme
+from adcp.compat.legacy.v2_5._url import extract_brand_domain
 
 
 def adapt_brand_manifest_to_brand(payload: dict[str, Any]) -> dict[str, Any]:
     """Rewrite v2.5 ``brand_manifest`` (URL string) to v3 ``brand``
-    ``{domain: ...}``. Caller-supplied ``brand`` wins when both fields
-    are present (half-migrated buyer)."""
+    ``{domain: ...}``.  Caller-supplied ``brand`` wins when both fields
+    are present (half-migrated buyer).
+
+    Uses ``extract_brand_domain`` to isolate the hostname from full URLs
+    (e.g. ``"https://acme.com/.well-known/brand.json"`` → ``"acme.com"``)
+    so the result satisfies ``BrandReference.domain``'s hostname-only regex.
+    """
     out = dict(payload)
     manifest = out.pop("brand_manifest", None)
     if isinstance(manifest, str) and manifest and "brand" not in out:
-        out["brand"] = {"domain": strip_url_scheme(manifest)}
+        out["brand"] = {"domain": extract_brand_domain(manifest)}
     return out
 
 
--- a/src/adcp/compat/legacy/v2_5/_url.py
+++ b/src/adcp/compat/legacy/v2_5/_url.py
@@ -8,6 +8,8 @@ import the same canonical implementation.
 
 from __future__ import annotations
 
+from urllib.parse import urlparse
+
 
 def strip_url_scheme(url: str) -> str:
     """``https://acme.example.com/`` → ``acme.example.com``.
@@ -16,6 +18,10 @@ def strip_url_scheme(url: str) -> str:
     after trailing-slash strip), ``http://`` schemes (legacy clients
     don't all enforce https), and trailing slashes from sloppy
     concatenation.
+
+    Does NOT strip URL paths or ports.  Use ``extract_brand_domain``
+    when the input may be a full URL (scheme + path) and you need only
+    the hostname component.
     """
     s = url.strip()
     for prefix in ("https://", "http://"):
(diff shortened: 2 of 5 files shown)


bokelley:
Promoting out of draft — CI green across 3.10–3.13, both expert reviews approved, regression tests in place for both `get_products` and `create_media_buy` adapter call sites.

Follow-up nits (not blocking this PR):

1. **`strip_url_scheme` naming.** Now that `extract_brand_domain` is the canonical helper for "URL → hostname", `strip_url_scheme` is reduced to the bare-domain fallback path. The name is misleading for the role it plays today; worth renaming or inlining in a follow-up.
2. **IPv6 silent failure.** `urlparse("https://[::1]/path").hostname` returns `"::1"`, which fails `BrandReference.domain`'s regex without a clear error. Acceptable for now (brand_manifest URLs in the wild are hostname-based, not IPv6 literals), but if we ever see this in production we should validate the result and raise a typed error rather than punting to the regex.

Tracking both as ideas; happy to file follow-up issues if useful.

bokelley:
Thanks for the promotion and the CI confirmation. Both nits are worth tracking — please do file follow-up issues for them. Brief suggested scopes:

1. **`strip_url_scheme` rename** — `_strip_url_scheme` (or inline into `extract_brand_domain`) to match its reduced, internal role.
2. **IPv6 validated error** — when `urlparse().hostname` returns a bare IPv6 literal, validate and raise a typed error rather than letting it fall through to the regex.

Neither is blocking; filing them now gives the next triage sweep a handle to act on them.

---
_Triaged by Claude Code._

---

bokelley:
## Expert review summary

**code-reviewer:** approved — no blockers. Confirmed ``urlparse().hostname`` is the right primitive; tests cover regression at both call sites. Minor nits: a bare-host-with-path (``acme.com/path``, no scheme) falls into ``strip_url_scheme`` and round-trips unchanged — fails downstream regex, which is acceptable fail-fast behaviour for an out-of-spec input.

**ad-tech-protocol-expert:** approved — confirmed the v2.5→v3 mapping is the only one v3's ``BrandReference`` schema permits (``additionalProperties: false``, domain-only). Surfaced two real spec-level follow-ups (non-blocking, filed as separate issues):

- #683 — v2.5 manifests at non-``/.well-known/brand.json`` paths get silently flattened to ``{domain}``; v3 sellers will fetch the canonical path and 404. Spec-level information loss the adapter cannot avoid; worth a log-warning.
- #684 — v2.5 ``brand_manifest`` ``oneOf[object, URL]`` is real; the adapter only handles the URL branch. Pre-existing gap surfaced during review.

CI: 16/16 green. Merging.

PR #680, opened 2026-05-11
fix(server): preserve Starlette lifespan when public_url is callable

### LINKED ISSUE #676: serve(transport="both") + callable public_url crashes at startup: 'function' object has no attribute 'router'
## Repro

```python
def my_resolver(request):
    return f"https://{request.headers.get('host')}/"

serve(handler, transport="both", public_url=my_resolver)
```

## Observed

Process exits cleanly with code 0 at lifespan startup; uvicorn logs:

```
ERROR:    Application startup failed. Exiting.
AttributeError: 'function' object has no attribute 'router'
```

Full traceback:

```
File "/.venv/.../adcp/server/serve.py", line 1705, in _composed_lifespan
    async with a2a_inner.router.lifespan_context(a2a_inner):
                ^^^^^^^^^^^^^^^^
AttributeError: 'function' object has no attribute 'router'
```

## Root cause

In `_serve_mcp_and_a2a` (`adcp/server/serve.py`), the lifespan composer at lines 1702–1706 assumes both `mcp_inner` and `a2a_inner` are Starlette apps with `.router`:

```python
async def _composed_lifespan(_app):
    async with mcp_inner.router.lifespan_context(mcp_inner):
        async with a2a_inner.router.lifespan_context(a2a_inner):
            yield
```

But `a2a_inner` is assigned from `create_a2a_server(..., public_url=public_url)` (line 1672). When `public_url` is a callable, `create_a2a_server` returns the `_wrap_with_per_request_card` middleware function (an `async def _middleware(scope, receive, send): ...`), not a Starlette app. The middleware has no `.router`.

The string and `None` cases pass `create_a2a_server`'s "use the static URL or the bound localhost" branch and return a Starlette app — that's why `transport="a2a"` alone works (it doesn't go through `_composed_lifespan`) and `transport="both"` works with non-callable `public_url`.

## Impact

Adopters following the docstring's per-tenant `PublicUrlResolver` pattern (parity with JS `createTenantRegistry`) on the unified MCP+A2A binary can't actually use the feature. Drove ~half a day of debugging on our side, including a CI run where every E2E and quickstart job timed out on "Server did not become healthy in time" because the container exited 0 instead of crashing loud.

## Suggested fix

Either:

1. **Have `_wrap_with_per_request_card` wrap the inner app instead of replacing it.** Keep `a2a_inner` as the Starlette app `create_a2a_server` would have built with `public_url=None`, then layer the per-request handler as middleware. The lifespan composer keeps working unchanged.

2. **Or make `_composed_lifespan` reach for the underlying app.** When `a2a_inner` doesn't have `.router`, walk a known attribute (`.app`, or store the inner reference on the wrapper).

Option 1 keeps the lifespan composer trivial; option 2 is a local fix.

## Workaround

Pass `public_url` as a string (e.g. from `PUBLIC_URL` env) and run your own ASGI middleware to rewrite `/.well-known/agent-card.json` from `X-Forwarded-Host` — same as the pre-5.1 pattern. We restored our 189-line `AgentCardPublicUrlMiddleware` for exactly this.

## Environment

- `adcp==5.2.0`
- Python 3.12
- `transport="both"`, multi-tenant subdomain deployment

bokelley:
## Summary

Fixes #676. ``serve(transport="both", public_url=<callable>)`` crashed at lifespan startup with ``AttributeError: 'function' object has no attribute 'router'`` and the process exited 0 with no requests served.

Root cause: ``create_a2a_server`` returned a raw ASGI callable (the per-request card middleware) when ``public_url`` was a ``PublicUrlResolver``, but the unified-transport composer in ``_serve_mcp_and_a2a`` reaches ``a2a_inner.router.lifespan_context`` directly. Only the static-string and ``None`` paths returned a Starlette app.

This implements **option 1 from the issue**: install the per-request card intercept as a proper Starlette middleware (``app.add_middleware(_PerRequestCardMiddleware, ...)``) so the returned object stays a Starlette instance with ``.router`` intact. The composer keeps working unchanged.

## Changes

- ``src/adcp/server/a2a_server.py`` — replaced ``_wrap_with_per_request_card`` closure with a class-based ASGI middleware ``_PerRequestCardMiddleware``; ``create_a2a_server`` now installs it via ``app.add_middleware(...)`` instead of wrapping. The card-serving behaviour is unchanged.
- ``tests/test_unified_mcp_a2a.py`` — added ``test_unified_app_starts_with_callable_public_url`` regression test. Verified it fails on the pre-fix source with the same ``AttributeError`` reported in the issue, and passes on the fix.

## Test plan

- [x] ``pytest tests/test_a2a_public_url_resolver.py tests/test_unified_mcp_a2a.py`` — 24 passed
- [x] Smoke: ``pytest tests/test_auth_middleware.py tests/test_mcp_middleware_composition.py tests/test_discovery_endpoint.py`` — 69 passed
- [x] ``ruff check`` clean on modified files
- [x] ``mypy src/adcp/`` clean
- [x] Black formatted (pre-commit hook re-ran)
- [x] Regression test verified to fail without the fix

### DIFF
--- a/src/adcp/server/a2a_server.py
+++ b/src/adcp/server/a2a_server.py
@@ -810,50 +810,63 @@ def _validate_card_url(url: str) -> str:
     return url
 
 
-def _wrap_with_per_request_card(
-    inner: Any,
-    *,
-    resolver: PublicUrlResolver,
-    handler: ADCPHandler[Any],
-    name: str,
-    port: int,
-    description: str | None,
-    version: str,
-    extra_skills: list[pb.AgentSkill] | None,
-    advertise_all: bool,
-    push_notifications_supported: bool,
-    auth: BearerTokenAuth | None,
-) -> Any:
-    """Wrap an ASGI app to serve agent-card endpoints per-request.
+_CARD_PATHS: frozenset[str] = frozenset({"/.well-known/agent-card.json", "/.well-known/agent.json"})
+
+
+class _PerRequestCardMiddleware:
+    """ASGI middleware that serves agent-card endpoints per-request.
 
     Intercepts GET ``/.well-known/agent-card.json`` and
-    ``/.well-known/agent.json``; all other requests pass through to the
-    inner app unchanged.
+    ``/.well-known/agent.json``; all other scopes (including
+    ``lifespan``) pass through unchanged.
 
-    Used when :func:`create_a2a_server` receives a
-    :data:`PublicUrlResolver` callable — the a2a-sdk's
-    ``create_agent_card_routes`` bakes the card at construction time
-    and cannot surface per-request context.
+    Installed via :meth:`starlette.applications.Starlette.add_middleware`
+    so the wrapped object remains a Starlette app — its ``.router``
+    stays reachable for lifespan composition in
+    :func:`adcp.server.serve._serve_mcp_and_a2a`.
     """
-    import inspect
 
-    from a2a.server.routes.agent_card_routes import agent_card_to_dict  # type: ignore[attr-defined]
-    from starlette.requests import Request
-    from starlette.responses import JSONResponse
+    def __init__(
+        self,
+        app: Any,
+        *,
+        resolver: PublicUrlResolver,
+        handler: ADCPHandler[Any],
+        name: str,
+        port: int,
+        description: str | None,
+        version: str,
+        extra_skills: list[pb.AgentSkill] | None,
+        advertise_all: bool,
+        push_notifications_supported: bool,
+        auth: BearerTokenAuth | None,
+    ) -> None:
+        self.app = app
+        self.resolver = resolver
+        self.handler = handler
+        self.name = name
+        self.port = port
+        self.description = description
+        self.version = version
+        self.extra_skills = extra_skills
+        self.advertise_all = advertise_all
+        self.push_notifications_supported = push_notifications_supported
+        self.auth = auth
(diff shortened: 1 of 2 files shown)


bokelley:
## Expert review summary

**code-reviewer:** approved — no blockers. Starlette middleware semantics correct, lifespan composition correct (router-level lifespan bypasses middleware stack which is the right behavior), regression test exercises the bug path. Nits (non-blocking): hot-path imports inside ``__call__`` could be hoisted; test could assert exact card field name rather than the ``supportedInterfaces or supported_interfaces`` fallback.

**python-expert:** approved — no blockers. Confirmed ``A2ABearerAuthMiddleware`` ordering is preserved (auth outside Starlette, card middleware inside — well-known paths short-circuit auth via the existing ``_A2A_DISCOVERY_PATHS`` allowlist). ``add_middleware`` after init is safe before first request. Lifespan path through ``_dispatch -> parent -> _composed_lifespan`` correctly bypasses the auth wrapper too.

CI: 16/16 green across Python 3.10–3.13, conformance, storyboards.

Merging.

bokelley:
Thanks for the review summary, @bokelley. No action needed on my end — happy to see the experts aligned and CI clean. 

Triaged by Claude Code. Session: https://claude.ai/code/${CLAUDE_CODE_REMOTE_SESSION_ID}

---

Issue #683, opened 2026-05-12
v2.5 brand_manifest with non-standard path is silently flattened to domain only


bokelley:
## Background

Surfaced during expert review of #679 (PR #679 fixes #677 — see https://github.com/adcontextprotocol/adcp-client-python/pull/679).

The v2.5 ``brand_manifest`` field is documented as a URL to a JSON manifest, and the spec's own example points at a non-``/.well-known/brand.json`` path:

> ``https://cdn.acmecorp.com/brand-manifest.json``

(from ``schemas/cache/2.5/core/brand-manifest-ref.json:29``)

After #679, the v2.5→v3 adapter correctly extracts the hostname so the result satisfies v3's ``BrandReference.domain`` hostname regex. But v3's ``BrandReference`` schema is hostname-only (``additionalProperties: false`` at ``schemas/cache/3.0/core/brand-ref.json``), so v3 sellers derive the canonical manifest URL as ``https://{domain}/.well-known/brand.json``.

## Impact

A v2.5 buyer hosting their manifest at a CDN or custom path (e.g. ``https://cdn.acmecorp.com/brand-manifest.json``) gets translated to ``{domain: \"cdn.acmecorp.com\"}``, and the v3 seller will try to fetch ``https://cdn.acmecorp.com/.well-known/brand.json`` — which will typically 404.

The adapter cannot do better given the v3 schema constraint (no field to preserve the original URL), but the silent flattening leaves both the buyer and the seller debugging a 404 with no signal in the SDK.

## Suggested mitigation

Emit a log warning at adapter-call time when the input URL's path is neither empty nor ``/.well-known/brand.json``. Something like:

> ``brand_manifest at <url> uses a non-standard path; v3 sellers derive {domain}/.well-known/brand.json from BrandReference.domain. Manifest fetch may 404.``

Doesn't change behaviour, but surfaces the mapping loss to operators.

## Out of scope here

Adding a v3 ``BrandReference.manifest_url`` field to preserve the original URL — that's a v3 spec change and belongs in the ``adcontextprotocol/adcontextprotocol`` repo, not here.

bokelley:
## Triage

**Classification:** Bug
**Bucket(s):** client/compat (no matching label exists in repo — gap noted)
**Status:** drafting-pr
**Milestone:** omitted (no milestone in issue text or linked PR)

**What the experts said:**
- code-reviewer: approved — no blockers; double `urlparse` call and dedup-key normalization are nits
[...]

PR #685, opened 2026-05-12
fix(compat): handle inline BrandManifest object in v2.5 adapters

### LINKED ISSUE #684: v2.5 brand_manifest as inline object (not URL) is unhandled by legacy adapter
## Background

Surfaced during expert review of #679. Pre-existing gap, not introduced by that PR.

The v2.5 ``brand-manifest-ref`` schema (``schemas/cache/2.5/core/brand-manifest-ref.json``) defines ``brand_manifest`` as:

```
oneOf:
  - BrandManifest object   # inline object
  - URL string             # reference URL
```

Both branches are spec-compliant v2.5 input.

## Impact

The current adapter only handles the **string** branch:

``src/adcp/compat/legacy/v2_5/get_products.py:130``:
```python
brand_manifest = out.pop(\"brand_manifest\", None)
if isinstance(brand_manifest, str) and brand_manifest and \"brand\" not in out:
    out[\"brand\"] = {\"domain\": extract_brand_domain(brand_manifest)}
```

And the matching helper at ``src/adcp/compat/legacy/v2_5/_media_buy_helpers.py`` (used by ``create_media_buy``) has the same guard.

When a v2.5 buyer sends an inline ``brand_manifest`` *object*, the field is popped (so it won't survive into v3), but ``brand`` is never populated either — the request goes through with no brand at all, which then fails v3 validation if ``brand`` is required, or silently advertises no brand if it isn't.

## Suggested fix

Handle the dict branch — extract the hostname from the inline object's URL field (whatever the v2.5 ``BrandManifest`` schema names it), or if the inline object has no derivable hostname, raise a typed translation error so the buyer gets a clear ``INVALID_REQUEST[brand_manifest]`` rather than a confusing downstream regex failure or silent drop.

Both call sites need the fix:
- ``src/adcp/compat/legacy/v2_5/get_products.py:adapt_request``
- ``src/adcp/compat/legacy/v2_5/_media_buy_helpers.py:adapt_brand_manifest_to_brand`` (used by ``create_media_buy``)

bokelley:
Closes #684

## Summary

The v2.5 `brand-manifest-ref.json` schema defines `brand_manifest` as a `oneOf` — either a URL string **or** an inline `BrandManifest` object. Both adapters (`get_products` and `create_media_buy`) only handled the string branch. When a v2.5 buyer sent an inline object, the field was popped but `brand` was never set — causing either a silent no-brand request or a confusing v3 validation failure downstream.

This PR adds the dict branch at both call sites:

- **Inline object with `url`:** extract hostname via existing `extract_brand_domain()`, set `brand: {domain}`. Guard uses `url.strip()` to prevent whitespace-only strings producing `{"domain": ""}`.
- **Inline object without `url`:** silently omit `brand` and let v3 validation decide. The `BrandManifest` schema marks `url` as optional (only `name` is required), so `{"name": "Great Value"}` is spec-valid input that cannot yield a hostname — raising here would reject spec-valid payloads.
- **Caller-supplied `brand` wins** over either form of `brand_manifest` (same precedence as the existing string branch).

`update_media_buy` intentionally does not translate `brand_manifest` and is unaffected.

## What was tested

- `pytest tests/test_legacy_adapter_v2_5_get_products.py tests/test_legacy_adapter_v2_5_media_buy.py -v` — **52 passed** (8 new tests: inline-with-url, url-with-path, no-url-skips-brand, brand-wins-over-inline, at both call sites)
- `ruff check src/adcp/compat/legacy/v2_5/` — no issues
- `mypy src/adcp/compat/legacy/v2_5/ --ignore-missing-imports` — no issues

- **Asymmetric required-field behavior (protocol expert):** `brand` is required in v3 `create_media_buy` but optional in v3 `get_products`. When an inline object has no `url`, the adapter silently omits `brand` at both sites. For `get_products` this is genuinely safe; for `create_media_buy` the request will still fail downstream with a v3 validation error. This is a deliberate "let v3 validate" posture — see discussion in #684 — but is worth noting for reviewers.
- **Module-level docstrings** in `create_media_buy.py` and `get_products.py` still say "URL string" for `brand_manifest`; in-function comments were updated but the top-level module docstring wire-shape tables were not.
- **Mutation-safety test** for the dict branch not added (the `out = dict(payload)` shallow copy is read-only for `.get("url")`, so no actual mutation risk, but coverage gap exists).

## Pre-PR review

- code-reviewer: approved — no blockers; whitespace-only `url` guard issue (blocker candidate) fixed before PR creation via `url.strip()`; nits noted above
- ad-tech-protocol-expert: approved — non-breaking per spec; `oneOf` handling correct; silent-skip for no-url is the only defensible behavior for an adapter that cannot fabricate a hostname; nit on asymmetric required-field behavior between call sites noted above

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01Nz3E58qfSW4X6fdJngTxKt

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nz3E58qfSW4X6fdJngTxKt)_

### DIFF
--- a/src/adcp/compat/legacy/v2_5/_media_buy_helpers.py
+++ b/src/adcp/compat/legacy/v2_5/_media_buy_helpers.py
@@ -37,7 +37,7 @@ def adapt_brand_manifest_to_brand(payload: dict[str, Any]) -> dict[str, Any]:
         out["brand"] = {"domain": extract_brand_domain(manifest)}
     elif isinstance(manifest, dict) and "brand" not in out:
         url = manifest.get("url")
-        if isinstance(url, str) and url:
+        if isinstance(url, str) and url.strip():
             out["brand"] = {"domain": extract_brand_domain(url)}
     return out
 
--- a/src/adcp/compat/legacy/v2_5/_media_buy_helpers.py
+++ b/src/adcp/compat/legacy/v2_5/_media_buy_helpers.py
@@ -18,9 +18,14 @@ from adcp.compat.legacy.v2_5._url import extract_brand_domain
 
 
 def adapt_brand_manifest_to_brand(payload: dict[str, Any]) -> dict[str, Any]:
-    """Rewrite v2.5 ``brand_manifest`` (URL string) to v3 ``brand``
-    ``{domain: ...}``.  Caller-supplied ``brand`` wins when both fields
-    are present (half-migrated buyer).
+    """Rewrite v2.5 ``brand_manifest`` (URL string or inline BrandManifest
+    object) to v3 ``brand`` ``{domain: ...}``.  Caller-supplied ``brand``
+    wins when both fields are present (half-migrated buyer).
+
+    v2.5 ``brand-manifest-ref.json`` is a oneOf: either a URL string or an
+    inline BrandManifest object.  For the inline case, the ``url`` field is
+    optional; when absent there is no derivable hostname, so ``brand`` is
+    omitted and v3 validation decides whether to reject.
 
     Uses ``extract_brand_domain`` to isolate the hostname from full URLs
     (e.g. ``"https://acme.com/.well-known/brand.json"`` → ``"acme.com"``)
@@ -30,6 +35,10 @@ def adapt_brand_manifest_to_brand(payload: dict[str, Any]) -> dict[str, Any]:
     manifest = out.pop("brand_manifest", None)
     if isinstance(manifest, str) and manifest and "brand" not in out:
         out["brand"] = {"domain": extract_brand_domain(manifest)}
+    elif isinstance(manifest, dict) and "brand" not in out:
+        url = manifest.get("url")
+        if isinstance(url, str) and url:
+            out["brand"] = {"domain": extract_brand_domain(url)}
     return out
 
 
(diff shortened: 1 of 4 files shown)


bokelley:
## Independent expert review (post-triage)

**code-reviewer (fresh):** approved — no blockers. Confirmed:

- ``isinstance(url, str) and url.strip()`` guard correctly rejects None / non-strings / whitespace-only inputs
- Silent-skip on missing ``url`` is the right call — preserves spec-correct behaviour of letting v3 validation own the missing-brand decision
- Brand precedence preserved (caller-supplied ``brand`` wins regardless of manifest shape) and tested at both call sites
- 8 new tests, symmetric across both adapters, behaviour-focused

**Nit (non-blocking):** the four-line dict branch is copy-pasted across ``get_products`` and ``_media_buy_helpers``. Borderline extract candidate — would consolidate to ``_media_buy_helpers.adapt_brand_manifest_to_brand`` as the single implementation. Not worth blocking; can fold in if a third call site appears.

CI: 15/15 green. Merging.

**Coordination note:** PR #686 (issue #683) is concurrently in flight on the same function. Will rebase #686 on top of this and extend the warning to the inline-object branch for coherence.

PR #688, opened 2026-05-12
fix(compat): warn on non-standard brand_manifest path in inline-object branch

### LINKED ISSUE #687: Extend brand_manifest non-standard-path warning to inline-object branch
## Background

Surfaced during rebase of PR #686 (#683) on top of PR #685 (#684), and called out by ``ad-tech-protocol-expert`` review of #686.

After both PRs merge, the v2.5 → v3 ``brand_manifest`` adapter handles two input shapes:

1. **URL string** — gets the non-standard-path warning from #686.
2. **Inline ``BrandManifest`` object** with optional ``url`` field — translated to ``brand.domain`` via ``extract_brand_domain(url)``, but the non-standard-path warning is NOT applied.

The same information loss occurs in both cases: a non-canonical path on the inline-object's ``url`` field will be silently flattened to the hostname, and v3 sellers' canonical ``/.well-known/brand.json`` fetch will likely 404.

## Suggested fix

Extract the warning logic into a shared helper (e.g. ``warn_if_brand_manifest_path_lossy`` in ``src/adcp/compat/legacy/v2_5/_url.py`` alongside ``extract_brand_domain``) and call it from both branches in both adapters.

This also addresses the duplication nit code-reviewer raised on #686 (warning logic copy-pasted across ``get_products.py`` and ``_media_buy_helpers.py``).

## Affected call sites

- ``src/adcp/compat/legacy/v2_5/get_products.py:adapt_request`` — inline-object branch
- ``src/adcp/compat/legacy/v2_5/_media_buy_helpers.py:adapt_brand_manifest_to_brand`` — inline-object branch

## Out of scope

Bounded dedup cache (``_brand_manifest_path_warned`` is currently an unbounded ``set``) — separate follow-up worth filing if production cardinality ever becomes a concern.

bokelley:
## Summary

Fixes #687. After #685 (inline-object branch) and #686 (non-standard-path warning) landed, the warning was applied only to the URL-string branch. The v2.5 ``brand-manifest-ref`` schema allows ``brand_manifest`` to be either a URL string or an inline ``BrandManifest`` object — the inline object's ``url`` field carries the same path semantics, and the same information loss applies (v3 sellers fetch the canonical ``{domain}/.well-known/brand.json`` and will likely 404).

## Changes

- ``src/adcp/compat/legacy/v2_5/_url.py``: new ``warn_brand_manifest_path_lossy(manifest_url, domain)`` function plus the ``_STANDARD_BRAND_MANIFEST_PATHS`` constant and ``_brand_manifest_path_warned`` dedup set (moved from per-adapter modules). Logger is ``adcp.compat.legacy.v2_5._url``.
- ``src/adcp/compat/legacy/v2_5/get_products.py`` and ``_media_buy_helpers.py``: remove the per-module helpers; call ``warn_brand_manifest_path_lossy`` from both the string and inline-object branches.
- Tests updated for the new monkeypatch target and logger name. Four new regression tests cover the inline-object warning behaviour (CDN url warns, canonical url silent) on both adapters.

## Test plan

- [x] ``pytest tests/test_legacy_adapter_v2_5_*.py`` — 64 passed (was 60; +4 new dict-branch warning tests)
- [x] Verified the two new tests fail against pre-fix source with ``assert 0 == 1`` (no warning emitted)
- [x] ``ruff check`` and ``mypy src/adcp/compat/legacy/v2_5/`` clean
- [x] Black formatted (pre-commit re-ran)

## Logger-name change

Operators previously alerting on ``adcp.compat.legacy.v2_5.get_products`` / ``adcp.compat.legacy.v2_5._media_buy_helpers`` warnings will need to update the logger name to ``adcp.compat.legacy.v2_5._url``. The warning message text is unchanged, so grep-based alerting on "non-standard path" is unaffected. Acceptable shift since #686 landed only an hour ago and adoption of log-based alerting on this specific warning is unlikely.

### DIFF
--- a/src/adcp/compat/legacy/v2_5/_media_buy_helpers.py
+++ b/src/adcp/compat/legacy/v2_5/_media_buy_helpers.py
@@ -12,19 +12,12 @@ v3 → v2 for responses.
 
 from __future__ import annotations
 
-import logging
 from typing import Any
-from urllib.parse import urlparse
 
-from adcp.compat.legacy.v2_5._url import extract_brand_domain
-
-_logger = logging.getLogger(__name__)
-
-# Paths that v3 sellers reconstruct correctly — no information is lost.
-_STANDARD_BRAND_MANIFEST_PATHS = {"", "/", "/.well-known/brand.json"}
-
-# Per-URL dedup so high-RPS v2.5 buyers don't saturate the log pipeline.
-_brand_manifest_path_warned: set[str] = set()
+from adcp.compat.legacy.v2_5._url import (
+    extract_brand_domain,
+    warn_brand_manifest_path_lossy,
+)
 
 
 def adapt_brand_manifest_to_brand(payload: dict[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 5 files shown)


bokelley:
## Expert review

- Refactor preserves behaviour exactly (same three-clause guard, same message format, same dedup semantics, same WARNING level)
- Both call sites symmetric after refactor
- 4 new regression tests well-scoped
- Logger-name change is a real (small) breaking shift but blast radius is near-zero since #686 landed ~1h ago — fine as-is

**Non-blocking nits filed for future work:**
- Pytest autouse fixture for ``_brand_manifest_path_warned`` reset would prevent future test pollution (currently every test monkeypatches; one missed reset would flake)
- Optional cross-adapter dedup test would document the now-shared-set semantic across ``get_products`` and ``create_media_buy``

CI: 15/15 green. Merging.

PR #690, opened 2026-05-12
test(compat): autouse fixture + cross-adapter dedup test for brand_manifest warning

### LINKED ISSUE #689: test: autouse fixture for brand_manifest warning dedup + cross-adapter dedup coverage
## Background

Follow-up from #688 review (PR for #687). Two non-blocking nits flagged by ``code-reviewer``:

1. **Autouse fixture for ``_brand_manifest_path_warned`` reset.** Every existing test for the brand_manifest non-standard-path warning does the same ``monkeypatch.setattr(_url_mod, "_brand_manifest_path_warned", set())`` boilerplate. A future test that forgets the reset would silently inherit state from prior tests and flake intermittently. Centralizing as an autouse fixture eliminates the boilerplate and removes the failure mode.

2. **Cross-adapter dedup test.** After #687 centralized the dedup state in ``_url.py``, the dedup set is shared across ``get_products`` and ``create_media_buy``. That's the intended behaviour (same offending URL through any adapter only warns once per process), but no test documents the semantic. If someone refactors back to per-module state, nothing catches it.

## Scope

- ``tests/conftest.py``: add scoped autouse fixture that resets ``adcp.compat.legacy.v2_5._url._brand_manifest_path_warned`` between tests.
- Remove the now-redundant ``monkeypatch.setattr(_url_mod, "_brand_manifest_path_warned", set())`` calls from the existing 10 tests in ``tests/test_legacy_adapter_v2_5_get_products.py`` and ``tests/test_legacy_adapter_v2_5_media_buy.py``.
- Add one cross-adapter test asserting that the same CDN URL through ``get_products.adapt_request`` and ``create_media_buy.adapt_request`` warns exactly once.

## Out of scope

- Bounded dedup cache (separate concern; per-deployment cardinality is small so unlikely to matter in practice).
- ``strip_url_scheme`` rename + IPv6 silent failure (nits filed during #679 review; cosmetic at this point).

bokelley:
## Summary

Fixes #689. Test-hygiene follow-ups flagged on the #688 review.

- Autouse fixture in ``tests/conftest.py`` clears ``adcp.compat.legacy.v2_5._url._brand_manifest_path_warned`` between every test. The 10 existing warning tests drop their ``monkeypatch.setattr(...)`` boilerplate. A future test that forgets the reset can no longer flake on inherited state.
- New ``test_brand_manifest_dedup_is_cross_adapter`` documents the shared-state semantic introduced in #687 — same offending URL through ``get_products`` then ``create_media_buy`` warns exactly once per process. Catches a regression back to per-module dedup.

## Verification

- ``pytest tests/test_legacy_adapter_v2_5_*.py`` — 65 passed (was 64; +1 cross-adapter test, no net loss from the removed monkeypatch calls).
- Manually simulated the per-module regression (``_brand_manifest_path_warned.clear()`` between adapter calls): two warnings emitted, confirming the new test would catch that change.
- ``ruff check`` + ``mypy`` clean.

## Out of scope

Other nits flagged across this cluster but not addressed (intentional — small or theoretical):

- Bounded dedup cache (``_brand_manifest_path_warned`` is an unbounded ``set``). Per-deployment cardinality of distinct ``brand_manifest`` URLs is small in practice; revisit if a real growth signal appears.
- ``strip_url_scheme`` rename (now a bare-domain fallback for ``extract_brand_domain``) — cosmetic.
- IPv6 silent failure on ``extract_brand_domain`` (``[::1]`` → ``"::1"`` fails ``BrandReference.domain`` regex without typed error) — theoretical for ``brand_manifest`` URLs in practice.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -74,6 +74,22 @@ def _a2a_compat_send_and_aggregate(
     _a2a_compat_shim.patch_send_and_aggregate(monkeypatch)
 
 
+@pytest.fixture(autouse=True)
+def _reset_brand_manifest_path_warned() -> None:
+    """Reset the v2.5 ``brand_manifest`` non-standard-path warning dedup
+    between tests.
+
+    The dedup lives in module-level state on ``adcp.compat.legacy.v2_5._url``
+    and is shared across both legacy adapters. Without this reset, a test
+    that fires the warning would suppress a subsequent test's expected
+    warning, producing order-dependent flakes. Autouse means tests don't
+    have to remember the boilerplate.
+    """
+    from adcp.compat.legacy.v2_5 import _url as _url_mod
+
+    _url_mod._brand_manifest_path_warned.clear()
+
+
 _adapter_cache: dict[type | UnionType, TypeAdapter[Any]] = {}
 
 
--- a/tests/test_legacy_adapter_v2_5_get_products.py
+++ b/tests/test_legacy_adapter_v2_5_get_products.py
@@ -147,12 +147,9 @@ def test_brand_manifest_inline_object_brand_wins_when_both_present() -> None:
 _URL_LOGGER = "adcp.compat.legacy.v2_5._url"
 
 
-def test_brand_manifest_cdn_url_warns(caplog, monkeypatch) -> None:  # type: ignore[no-untyped-def]
+def test_brand_manifest_cdn_url_warns(caplog) -> None:  # type: ignore[no-untyped-def]
     """CDN brand_manifest with non-standard path emits a WARNING about the
     information loss so operators can debug downstream 404s."""
-    import adcp.compat.legacy.v2_5._url as _url_mod
-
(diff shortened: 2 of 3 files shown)


bokelley:
## Expert review

- ``.clear()`` vs ``monkeypatch.setattr(..., set())`` semantic change is safe (state is only referenced via attribute lookup inside ``_url.py``; no code caches the set by reference)
- Cross-adapter test is correctly placed with an explicit regression-mode assertion message
- Direct attribute access for the test-only reset is preferable to expanding the public API with a ``clear_dedup()`` helper

**Nit (non-blocking):** the autouse fixture lives at top-level ``conftest.py`` so it runs ``.clear()`` on every test in the suite. Cost is negligible; a scoped sub-conftest would be more surgical if compat tests ever get their own subdirectory.

CI: 15/15 green. Merging.

PR #691, opened 2026-05-12
refactor(compat): normalize + cap brand_manifest warning dedup

bokelley:
## Summary

Three follow-up nits raised across the brand_manifest cluster reviews (#679, #686, #688), bundled into one PR since all three touch ``src/adcp/compat/legacy/v2_5/_url.py``:

1. **``strip_url_scheme`` inlined** as a private ``_bare_domain`` fallback inside ``extract_brand_domain``. Single-call-site, name reflects current role.
2. **Dedup key normalized** to ``(scheme, netloc, path)`` with a single trailing-slash stripped from path. Cache-buster query-string variants collapse to one dedup slot; ``/.well-known/brand.json/`` is no longer a false positive on the standard-paths check.
3. **Dedup set capped** at 1024 with clear-on-overflow. Defense-in-depth against pathological URL cardinality; memory stays bounded.

IPv6 nit handled by docstring only — current behaviour (unsupported hostnames produce a clean ``BrandReference.domain`` regex-mismatch error from v3 Pydantic) is actually informative for buyers; changing to silent-skip or raise both have worse trade-offs.

## Test plan

- [x] ``pytest tests/test_legacy_adapter_v2_5_*.py`` — 68 passed (was 65; +3 new tests for cache-buster dedup, trailing-slash canonical, cap clear-on-overflow)
- [x] Broader smoke (``test_a2a_public_url_resolver.py``, ``test_unified_mcp_a2a.py``) — 92 passed total
- [x] ``ruff check`` and ``mypy src/adcp/compat/legacy/v2_5/`` clean

## Notes

- ``strip_url_scheme`` is removed from the public surface of ``_url.py``. It was never exported via ``__init__.py`` and ``rg`` confirms zero external callers (only internal use inside ``extract_brand_domain``). Safe removal.
- The dedup-key shape changed from ``str`` to ``tuple[str, str, str]``. The autouse fixture in ``tests/conftest.py`` calls ``.clear()`` and so is unaffected.

### DIFF
--- a/src/adcp/compat/legacy/v2_5/_url.py
+++ b/src/adcp/compat/legacy/v2_5/_url.py
@@ -15,26 +15,34 @@ _logger = logging.getLogger(__name__)
 
 # Paths that v3 sellers reconstruct correctly from ``brand.domain`` — no
 # information is lost when the adapter flattens these to the hostname.
-_STANDARD_BRAND_MANIFEST_PATHS = {"", "/", "/.well-known/brand.json"}
+# Compared after a single trailing slash is stripped from the input path,
+# so ``/.well-known/brand.json/`` is treated the same as the canonical form.
+_STANDARD_BRAND_MANIFEST_PATHS = {"", "/.well-known/brand.json"}
 
 # Per-URL dedup so high-RPS v2.5 buyers don't saturate the log pipeline
-# when many requests carry the same non-canonical manifest URL.
-_brand_manifest_path_warned: set[str] = set()
+# when many requests carry the same non-canonical manifest URL. The key is
+# normalized via ``urlparse`` (scheme + netloc + path-stripped-of-trailing-slash)
+# so cache-buster query strings (``?v=1``, ``?v=2`` …) collapse to one entry.
+_brand_manifest_path_warned: set[tuple[str, str, str]] = set()
+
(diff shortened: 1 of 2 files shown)


bokelley:
## Expert review

- Behaviour preserved across all 10 existing warning tests (normalized dedup key produces same outcome for the inputs they exercise — no query strings, no trailing-slash variants).
- ``strip_url_scheme`` removal safe (zero external callers, never re-exported).
- ``_STANDARD_BRAND_MANIFEST_PATHS`` dropping ``"/"`` is correct given ``rstrip("/")`` runs before the membership check; empty string covers both ``""`` and ``"/"``.
- Cap clear-on-overflow is fine for the stated threat model; a misbehaving buyer rotating >1024 paths per cycle produces duplicate warnings, acceptable per the PR rationale.
- IPv6 docstring-only approach is sound — "translate, don't reject" contract is the right call; Pydantic's pattern-mismatch error at the v3 boundary is more informative than a silent skip.
- Dedup key type migration (``set[str]`` → ``set[tuple[str, str, str]]``) is shape-agnostic from the test fixture's perspective.
- New overflow test correctly ``caplog.clear()``s between fill and trigger phases.

**Nit (non-blocking):** docstring could mention that operators seeing repeat warnings for the same URL should check whether they've crossed the dedup cap.

CI: 15/15 green. Merging.

PR #692, opened 2026-05-12
feat(buyer-agent-registry): with_caching factory auto-invalidates on mutations

### LINKED ISSUE #380: feat(buyer-agent-registry): add caching + rate-limiting + audit emission to PgBuyerAgentRegistry
## Context

`PgBuyerAgentRegistry` (and the v3 ref seller's `TenantScopedBuyerAgentRegistry` which delegates to it) hits the database on every dispatch with:

- no in-process cache
- no rate limit
- no audit emission on resolve outcomes

Production sellers will eat unnecessary DB load, and the lookup endpoint becomes a credential-stuffing oracle for an attacker who can probe it.

Refs:
- `src/adcp/decisioning/pg/buyer_agent_registry.py`
- `examples/v3_reference_seller/src/buyer_registry.py`

## Acceptance Criteria

- [ ] In-process TTL cache (default 60s, configurable) keyed on the lookup tuple
- [ ] Per-tenant (or per-source-IP) rate limit on the resolve path, configurable
- [ ] Audit event emitted on every resolve outcome (hit / miss / rate-limited / error) via the existing audit hook
- [ ] Metrics hooks (counter for hits/misses/rate-limits, latency histogram)
- [ ] Cache invalidation on registry mutations (sync / delete)
- [ ] Unit tests cover cache hit/miss/expiry, rate-limit triggering, audit emission

bokelley:
Closes #380.

## Why

Adopters wiring `Caching(Pg(...))` today must remember to call `cache.invalidate(...)` after every `pg.upsert / set_status / delete` or risk serving stale resolutions for up to `ttl_seconds`. That's a real security footgun — an operator suspending an agent expects the next dispatch to deny, not to honor a cached "active" entry.

## Wire-and-forget path

```python
pg = PgBuyerAgentRegistry(pool=pool)
registry = pg.with_caching(ttl_seconds=60)
serve(buyer_agent_registry=registry, ...)

# Cache auto-invalidates on mutations through pg:
pg.set_status("https://agent/", "suspended")
# Next resolve() through `registry` hits DB, sees suspended.
```

## Implementation

- `PgBuyerAgentRegistry.add_mutation_observer(callback)` registers a sync callback fired after every successful `upsert / set_status / delete`. Observer exceptions are logged and swallowed so a broken hook never blocks a real mutation.
- `PgBuyerAgentRegistry.with_caching(**cache_kwargs)` builds a `CachingBuyerAgentRegistry` and pre-registers a `clear_sync` observer on it.
- `CachingBuyerAgentRegistry.clear_sync()` — sync-callable, no event loop required, atomic via GIL on `dict.clear()`. Full-flush rather than per-key drop because mutations are admin-rare and the next traffic burst rebuilds the working set within TTL.

## Tradeoffs

- **Full-clear over per-key invalidation:** simpler code, no secondary index needed, no race window around the cache's `asyncio.Lock`. The over-invalidation is bounded — admin mutations happen at human pace, not request pace.
- **Same-instance scope:** the observer hook fires on mutations through *this* `PgBuyerAgentRegistry` instance only. Adopters with external admin paths (a separate process writing the same DB) still need `.invalidate(...)` or `.clear_sync()` explicitly. Documented in the `with_caching` docstring.

## What was tested

- `pytest tests/test_buyer_agent_registry_cache.py -q` — 25 passed (2 new tests for `clear_sync` + 23 existing)
- 4 new PG conformance tests in `tests/conformance/decisioning/test_pg_buyer_agent_registry.py` covering: observer on upsert, observer on set_status + delete, observer exception swallowing, end-to-end `pg.with_caching()` invalidation
- `ruff check src/adcp/decisioning/pg/buyer_agent_registry.py src/adcp/decisioning/registry_cache.py` — clean
- `mypy src/adcp/decisioning/pg/buyer_agent_registry.py src/adcp/decisioning/registry_cache.py --ignore-missing-imports` — clean

### DIFF
--- a/src/adcp/decisioning/pg/buyer_agent_registry.py
+++ b/src/adcp/decisioning/pg/buyer_agent_registry.py
@@ -79,7 +79,9 @@ from __future__ import annotations
 
 import asyncio
 import json
+import logging
 import re
+from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
 from adcp.decisioning.registry import (
@@ -92,6 +94,17 @@ from adcp.decisioning.registry import (
 
 if TYPE_CHECKING:
     from psycopg_pool import ConnectionPool
+
+    from adcp.decisioning.registry_cache import CachingBuyerAgentRegistry
+
+logger = logging.getLogger(__name__)
+
+MutationObserver = Callable[[str, str], None]
+"""Callback fired after a successful :class:`PgBuyerAgentRegistry`
+mutation. Receives ``(operation, agent_url)`` where ``operation`` is
+``"upsert"`` / ``"set_status"`` / ``"delete"``. Runs synchronously on
+the caller's thread after the DB commit; raised exceptions are logged
+but do not propagate to the mutation caller."""
 
 try:
     import psycopg  # noqa: F401
@@ -157,6 +170,7 @@ class PgBuyerAgentRegistry:
             )
         self._pool = pool
         self._table = table_name
+        self._mutation_observers: list[MutationObserver] = []
 
         # Pre-format queries so the hot path doesn't f-string per call.
         # All identifier substitutions are validated at __init__; row
@@ -301,6 +315,7 @@ class PgBuyerAgentRegistry:
         )
         with self._pool.connection() as conn, conn.cursor() as cur:
             cur.execute(self._sql_upsert, params)
+        self._notify_mutation("upsert", agent.agent_url)
 
     def set_status(self, agent_url: str, status: BuyerAgentStatus) -> None:
         """Update an agent's lifecycle status. Use to suspend / block
@@ -309,6 +324,7 @@ class PgBuyerAgentRegistry:
             raise ValueError(f"status must be one of {sorted(_VALID_STATUSES)!r}, got {status!r}")
         with self._pool.connection() as conn, conn.cursor() as cur:
             cur.execute(self._sql_set_status, (status, agent_url))
+        self._notify_mutation("set_status", agent_url)
 
     def delete(self, agent_url: str) -> None:
         """Remove an agent from the registry.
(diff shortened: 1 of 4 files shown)


bokelley:
## Review

CI is green across all 13 Python + storyboard jobs, conformance tests cover the four obligations called out in the PR body. Recommending merge — flagging a few non-blocking observations for follow-up.

### Solid

- The wire-and-forget shape (`pg.with_caching(ttl_seconds=...)`) collapses what was a four-step incantation. This is the right call — the manual-invalidation path was a known footgun and several adopters were going to hit it eventually.
- Observer hook fires *after* the `with self._pool.connection() as conn:` block exits, so committed mutations and committed mutations only invalidate. The exception-swallowing is correct: an observer crash must not roll back a real DB mutation.
- Full-flush over per-key invalidation is the right tradeoff at admin-mutation cadence. The race-window analysis in the `clear_sync` docstring is correct (worst case: one extra DB round-trip, which is exactly what an invalidation should cause).

### Non-blocking observations

1. **Observer registry is append-only + unlocked.** `self._mutation_observers: list[MutationObserver]` has no remove method and `add_mutation_observer` writes without a lock. Mutation hot-path iterates the list. In practice observers are wired at boot and stay forever, so the missed-write race is theoretical — but tests that wire fixtures across registry instances will accumulate observers if the registry is reused. Worth a `remove_mutation_observer` for symmetry. Will file as follow-up.

2. **`with_caching` only wires the cache layer.** The module docstring in `registry_cache.py:1-30` documents the canonical stack as `Caching(RateLimited(Auditing(inner)))`. Adopters who want the full stack still have to construct it manually and then call `pg.add_mutation_observer(...)` themselves. A `pg.with_full_stack(ttl_seconds=..., rps_per_tenant=..., audit_sink=...)` would close the loop but is genuinely more opinionated. Will file as follow-up, not block this.

3. **`clear_sync` docstring claim about GIL atomicity is true for CPython 3.13 and earlier.** Under PEP 703 free-threaded CPython (3.13t / 3.14t) the `dict.clear()` is still atomic per-object but the "no lock acquired" reasoning leans on a CPython implementation detail. Worst case is still bounded over-invalidation — fine to keep as written, but worth a one-line note for adopters running on `--disable-gil` builds. Not blocking.

4. **Docstring nit:** the "Limitations" line in `with_caching` mentions external admin paths (separate process writing the DB) but doesn't call out the more common case — adopters who hold both a raw `pg` reference *and* the cache, mutate through a *second* `pg` instance pointed at the same pool. Same root cause, different shape, and easier to do by accident than the "separate process" framing suggests.

LGTM modulo the follow-ups (which I'll file separately).

Issue #377, opened 2026-05-03
feat(v3-ref-seller): implement sync_accounts + list_accounts with billing_entity write-only projection guard wired


bokelley:
## Context

The v3 reference seller has the `Account.billing_entity` JSON column but does not implement `sync_accounts` or `list_accounts`. Because of this, `adcp.types.project_account_for_response` (the 3.1 bank-details write-only guard) is never called — the protection is purely architectural, not enforced.

Refs:
- `examples/v3_reference_seller/src/platform.py`
- `examples/v3_reference_seller/src/models.py` (`Account.billing_entity`)
- `adcp.types.project_account_for_response`

## Acceptance Criteria

- [ ] `sync_accounts` implemented — persists incoming accounts (including `billing_entity` with bank details), then returns response via `project_account_for_response`
- [ ] `list_accounts` implemented — returns accounts projected through `project_account_for_response` so bank details are stripped
- [ ] Smoke test asserts `bank_details` (or any write-only field) never appears in either response
- [ ] Smoke test asserts non-write-only fields (display name, address, tax id format) DO appear

bokelley:
## Triage

**Classification:** Feature request
**Bucket(s):** examples
**Status:** drafting-pr
**Milestone:** (omitted — no milestone on file)

**What the experts said:**
- dx-expert: Non-blocking feature; natural key derivation needs adopter-customization notice (fixed); dry_run comment needs SQLAlchemy rollback sketch (fixed); blocker on bank-stripping helper choice explanation (fixed)
- docs-expert: Mock-based tests consistent with existing no-PG philosophy; assert on `model_dump(mode="json")` not raw Pydantic attr; module docstring should list new methods (done)
- code-reviewer (pre-PR): Three blockers fixed — missing `name` in sync response dict; `sandbox` None→False inconsistency between DB and wire; `dry_run=True` echoed back when session committed (lie on wire, fixed to always return `False`)

**My take:** Straightforward non-breaking addition to the examples bucket: two methods + two smoke tests, all wired to the already-merged write-only guard. Draft PR #403 covers all four acceptance criteria.

**PR:** #403 — `feat(examples): sync_accounts + list_accounts with billing_entity write-only guard wired`

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01N51oUL2dxvvaVqe2xcjTuf

---
_Generated by [Claude Code](https://claude.ai/code)_

bokelley:
Closed by PR #693.

[...]

PR #693, opened 2026-05-12
fix(ci): v3 reference seller storyboard job actually asserts on results

### LINKED ISSUE #410: feat(testing): storyboard runner matrix entry for examples/v3_reference_seller
## Context

The CI storyboard runner currently targets `examples/seller_agent.py` only. The v3 reference seller (`examples/v3_reference_seller/src/app.py`) now has all 9 sales-non-guaranteed methods, projection guard, and traffic counters (per #405) — but ships unverified at the storyboard level. End-to-end coverage is missing for the seller adopters are most likely to fork.

The Postgres service required by the v3 ref seller is already provisioned for the `postgres-conformance` job, so the same service can be reused.

## Acceptance Criteria

- [ ] Dedicated GHA matrix job that boots `examples/v3_reference_seller/src/app.py`
- [ ] Storyboard runner executes against the booted seller
- [ ] Asserts at least one `sales-non-guaranteed` storyboard passes end-to-end
- [ ] Asserts `/_debug/traffic` shows the expected per-method counts (the anti-façade contract from PR #405)
- [ ] Reuses Postgres service from `postgres-conformance` job

## Refs

- PR #405 (traffic counters / MockAdServer Protocol)
- PR #408 (broadened sales surface + sync_accounts + invoice_recipient)
- PR #402 (account.supported_billing)

bokelley:
Closes #410.

## Problem

The \`storyboard-v3-reference-seller\` job advertised itself as gating "every PR's translator-pattern conformance" but ended its storyboard invocation with \`|| true\` (`.github/workflows/ci.yml:645`), so a failing run silently passed CI. The artifact path also pointed at \`examples/v3_reference_seller/\` while the file was written to workspace root, so post-mortem inspection landed on \`if-no-files-found: warn\` instead of the actual result.

The job had cosmetic gate-value, no real gate-value.

## Fix

Three changes:

1. **Drop \`|| true\`** from the storyboard invocation. \`cd\` into \`examples/v3_reference_seller\` so the result file lands where \`upload-artifact\` expects it.
2. **"Assert storyboard passed"** step matching the seller_agent job's pattern (`ci.yml:435-451`): \`overall_status == 'passing'\` AND \`controller_detected == true\`. On failure, dump the full JSON so the diagnostic IS the storyboard report.
3. **"Assert upstream traffic" anti-façade gate**: GET \`/_debug/traffic\` from the seller; fail if all per-method counts are zero. A seller that returns stub data without translating would have passed the storyboard but emitted no traffic — this is the failure mode #410 specifically called out.

## What was tested

- \`python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"\` — YAML valid
- CI will run on this PR; the v3 reference seller already implements \`enable_debug_endpoints=True\` (per `examples/v3_reference_seller/src/app.py:192`) so the \`/_debug/traffic\` endpoint exists and will return real counts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -676,48 +676,19 @@ jobs:
           print('summary:', json.dumps(d.get('summary', {}), indent=2))
           "
 
-      - name: Assert storyboard passed (allowlist)
-        # Hard gate with an explicit allowlist of expected failures. Any
-        # failure NOT in the allowlist fails the job; any allowlisted
-        # failure that newly PASSES also fails the job (so the list stays
-        # honest and gets pruned as upstream issues close).
-        #
-        # Allowlist:
-        #   - media_buy_seller/refine_products/get_products_refine — the
-        #     v3 reference seller does not implement refine_get_products();
-        #     the framework correctly returns INVALID_REQUEST for refine
-        #     requests on platforms without that optional method. Feature
-        #     gap tracked separately.
+      - name: Assert storyboard passed
+        # Hard gate. The runner's ``overall_status`` MUST be ``passing``;
+        # any failure fails the job.
         run: |
           python -c "
           import json, sys, pathlib
-          allowed = {
-              ('media_buy_seller/refine_products', 'get_products_refine'),
-          }
           p = pathlib.Path('examples/v3_reference_seller/v3-storyboard-result.json')
           with p.open() as f:
               d = json.load(f)
-          failures = d.get('failures') or []
-          observed = {(f.get('storyboard_id'), f.get('step_id')) for f in failures}
-          unexpected = observed - allowed
-          fixed = allowed - observed
-          ok = True
-          if unexpected:
-              print('UNEXPECTED storyboard failures (not in allowlist):')
-              for sid, step in sorted(unexpected):
-                  print(f'  - {sid} / {step}')
-              ok = False
-          if fixed and d.get('overall_status') == 'passing':
-              # All allowlisted failures resolved AND overall is passing —
-              # drop the allowlist entirely.
-              print('All allowlisted failures are now passing. Remove this allowlist step.')
-              ok = False
-          if not ok:
-              print()
-              print('Full result follows:')
+          if d.get('overall_status') != 'passing':
               print(json.dumps(d, indent=2))
               sys.exit(1)
-          print(f'Storyboard residuals match allowlist ({len(observed)} expected).')
+          print('Storyboard passing.')
           "
 
       - name: Assert upstream traffic (anti-façade gate)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -683,19 +683,16 @@ jobs:
         # honest and gets pruned as upstream issues close).
         #
         # Allowlist:
-        #   - media_buy_seller/refine_products/get_products_refine — upstream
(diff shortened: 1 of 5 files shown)


bokelley:
## Review

The patch itself is right — drop `|| true`, assert `overall_status == 'passing'`, assert non-zero upstream traffic. That's exactly the gate description.

**But the gate is already doing its job: the new \`AdCP storyboard runner — v3 reference seller (translator)\` check is failing on this PR with 12 storyboard-level failures**, which is what the previous `|| true` was masking. Three distinct failure clusters:

### 1. Storyboard fixture / authorization setup (7 failures — same root cause)

All 7 `media_buy_seller/*/get_products_brief` steps fail with:

> `PERMISSION_DENIED: Buyer agent is not authorized for this seller. The seller's commercial allowlist did not authorize this credential.`

This looks like a storyboard buyer-agent fixture issue — the runner is presenting credentials the v3 ref seller's commercial allowlist doesn't carry. Either the storyboard scenarios assume a fixture step that no longer runs, or the v3 ref seller's seed data missed the buyer-agent registry. **Worth confirming locally before landing this PR** — if it's a fixture gap on the reference seller side, fix in same PR; if it's a real allowlist policy regression, file separately.

### 2. Cascade-rule error code mismatches (2 failures)

- `create_media_buy_aggressive_terms`: expected `TERMS_REJECTED`, got something else at `/adcp_error/code`
- `update_unknown_media_buy`: expected `MEDIA_BUY_NOT_FOUND`
- `update_unknown_package`: expected `PACKAGE_NOT_FOUND`

These are real spec-conformance regressions in the v3 ref seller. Each one should be a separate issue; the cascade-rules layer can't normalize these because they're terminal codes.

### 3. Spec conformance: missing `list_accounts` / `sync_accounts` (1 failure)

```
Agent declared account-bearing specialism(s) [sales-non-guaranteed, sales-guaranteed]
but advertises neither list_accounts nor sync_accounts.
```

This is **exactly #377**, which is already open. Worth linking that issue from this PR's description so the dependency is explicit.

### 4. Storyboard step bug?

> `INVALID_REQUEST: buying_mode='refine' must not be combined with brief`

`get_products_refine` storyboard step is sending `buying_mode=refine` alongside `brief`. Either the spec changed (refine doesn't take brief), or the storyboard step is wrong. Verify against the spec slice.

### Process recommendation

The PR has correctly diagnosed and fixed the cosmetic-gate problem. The trade-off now is:

- **Land as-is** → the v3 storyboard job goes red on `main` for every PR until the 12 failures are fixed. This is the honest signal but blocks unrelated work.
- **Hold** until the 12 failures are fixed (or at least the fixture issue, which is 7 of them) → defeats the purpose somewhat but keeps `main` green.
- **Land + immediately file follow-ups** (preferred) — capture each cluster as a tracked issue, link from PR body, and accept the temporary red. The whole point of the gate is that broken conformance should be visible, not hidden.

Also: branch is `BEHIND` base. Needs rebase before merge regardless.

I'll file follow-ups for clusters 2 and 4 separately. Cluster 1 needs human eyes on the v3 ref seller fixture setup — please confirm before I file.

bokelley:
Filed the PERMISSION_DENIED cluster as #703 with three hypotheses (fixture seed gap / allowlist policy regression / middleware drift) and triage steps. Team can pick it up from there.

Full follow-up set for this PR's failures: #701, #702, #703 + already-open #377.

Issue #678, opened 2026-05-11
feat(testing): in-process A2A test client (sibling to SellerTestClient)


bokelley:
Parent: #662
MCP sibling shipped in #666 (`SellerTestClient.invoke()`).

## Why

`SellerTestClient` eliminates JSON-RPC envelope + `adcp_error` extraction boilerplate for MCP adopters. A2A adopters need the same harness shape — but A2A's semantics are different enough that a shared `invoke()` would force-fit both.

## What's different about A2A

A2A is task-state-machine, not call/return:

- `task.create(method, payload)` → returns a task handle in `submitted` / `working` state
- State transitions through `TaskState.TASK_STATE_*` (see `src/adcp/server/a2a_server.py:467,482,547`)
- Terminal states: `COMPLETED` / `FAILED` / `CANCELED`
- Async completion is delivered via push-notification webhooks the seller signs

A `SellerTestClient.invoke()` shape collapses all of that into a single call, hiding exactly the lifecycle adopters need to test.

## Proposed shape (not final — open to triage)

```python
from adcp.testing import SellerA2AClient   # name TBD

async def test_create_buy_completes(seller_a2a):
    handle = await seller_a2a.create_task(
        "create_media_buy", {"products": [...], ...}
    )
    result = await handle.wait_for(state="completed", timeout=5.0)
    assert result.passed
    assert result.data[\"media_buy_id\"]

async def test_push_notification_signed_on_completion(seller_a2a):
    handle = await seller_a2a.create_task("create_media_buy", {...})
    notif = await seller_a2a.expect_push_notification(timeout=5.0)
    assert notif.task_id == handle.task_id
    assert notif.signed_with_kid == \"kid_test\"
    assert notif.body[\"final\"] is True
```

Surfaces the harness likely needs:

- `create_task(method, payload) → TaskHandle`
- `TaskHandle.wait_for(state=..., timeout=...) → ToolInvokeResult`
- `TaskHandle.cancel()`
- An in-process push-notification sink (captures outbound webhooks instead of HTTP-POSTing them) so adopters can assert on signing, body shape, and ordering

## Why this matters for the broader A2A adoption story

A2A adoption is currently deferred behind \"pluggable TaskStore + push-notification + middleware hooks.\" An in-process test harness gives adopters something concrete to write against once those hooks land — and exercises the hooks themselves, which makes the harness a useful design forcing-function.

## Scope notes

[...]

PR #666, opened 2026-05-11
feat(testing): add SellerTestClient for in-process handler testing

bokelley:
Refs #662

Adds `SellerTestClient` — an in-process MCP test harness that eliminates the per-file boilerplate of constructing JSON-RPC envelopes, calling `mcp.call_tool()`, and extracting `adcp_error` from `structuredContent`. Adopters now get a single typed call that handles all three.

```python
from adcp.testing import SellerTestClient

@pytest.fixture
def seller():
    return SellerTestClient(MySeller())

async def test_buy_not_found(seller):
    result = await seller.invoke("update_media_buy", {"media_buy_id": "missing", ...})
    assert not result.passed
    assert result.adcp_error.code == "MEDIA_BUY_NOT_FOUND"
```

The harness calls `mcp.call_tool()` in-process — it exercises the full handler dispatch and error-translation stack without HTTP or SSE framing. For HTTP-level tests (auth, CORS, middleware), `build_test_client` remains the right tool.

**What this PR does not include** (see #662 for status):
- `run_scenario()` — requires bundled compliance scenario playbooks that are not yet in the SDK
- A2A transport — served via a separate ASGI app; needs its own in-process client primitive

## What was tested

- `pytest tests/test_seller_test_client.py` — 18 new tests, all passing
- `pytest tests/ -q --ignore=tests/integration` — 4447 passed, 1 pre-existing network flake (`test_real_tls_handshake_still_validates_hostname`)
- `ruff check src/adcp/testing/` — no issues
- `mypy src/adcp/testing/harness.py --ignore-missing-imports` — no new errors from this file

## New public API

| Symbol | Module | Notes |
|---|---|---|
| `SellerTestClient` | `adcp.testing` | In-process MCP harness; takes `DecisioningPlatform` |
| `ToolInvokeResult` | `adcp.testing` | Return type with `.passed`, `.data`, `.adcp_error`, `.raw` |
| `AdcpErrorPayload` | `adcp.testing` | Typed container for wire `adcp_error` fields |

`validation=None` (default) disables schema validation so tests target handler behavior. Pass `SERVER_DEFAULT_VALIDATION` to match production behavior (`build_test_client` does this by default — see migration note in `__init__` docstring).

## Nits surfaced by review (not fixed in this PR)

- Private `_mcp` attribute accessed in two tests — low-value coupling to implementation detail; safe to remove in a follow-up cleanup
- `docs/testing-guide.md` has no mention of `SellerTestClient` — should be added once the API stabilises
[...]

PR #694, opened 2026-05-12
feat(testing): add SellerA2AClient for in-process A2A handler testing

bokelley:
Refs #678.

## Why

\`SellerTestClient\` shipped in #666 killed the JSON-RPC envelope + \`structuredContent\` extraction boilerplate every MCP-adopting test file was rewriting by hand. Adopters who ship A2A in production today (salesagent, the v3 reference seller in this SDK) have been rewriting the exact same shape of boilerplate against the A2A executor — building proto Messages with DataParts, plumbing a SendMessageRequest into RequestContext, draining an EventQueueLegacy, projecting the terminal Task's first artifact back to a dict, and handling structured vs. unstructured failure modes separately.

\`SellerA2AClient\` collapses that into the same one-call shape as the MCP version.

## What this PR ships

\`\`\`python
from adcp.testing import SellerA2AClient

@pytest.fixture
def seller_a2a():
    return SellerA2AClient(MySeller())

async def test_buy_not_found(seller_a2a):
    result = await seller_a2a.invoke(
        "update_media_buy", {"media_buy_id": "missing", ...}
    )
    assert not result.ok
    assert result.adcp_error.code == "MEDIA_BUY_NOT_FOUND"
\`\`\`

- New \`SellerA2AClient\` in \`adcp.testing\`. Same constructor shape as \`SellerTestClient\` (takes a \`DecisioningPlatform\`, optional \`validation\` config).
- Reuses \`ToolInvokeResult\` and \`AdcpErrorPayload\` — assertion grammar is identical across MCP and A2A test suites.
- Drains the A2A event queue with a bounded loop + per-event timeout (default 5s). A buggy handler that never publishes a terminal event raises a clear \`RuntimeError\` instead of hanging the test runner.
- Structured A2A errors (AdcpError raised by handler, \`adcp_error\` in failed-Task DataPart per transport-errors.mdx §A2A Binding) extract identically to MCP.
- **Unstructured A2A errors** (unknown skill, unparseable message — FAILED Task with no \`adcp_error\` payload) synthesize into an \`INTERNAL_ERROR\`-coded \`AdcpErrorPayload\` so callers can assert on \`result.adcp_error\` uniformly across structured/unstructured failures.

## What this PR does NOT ship (deferred to follow-ups on #678)

These are the genuinely A2A-shaped concerns that motivated the original issue framing:

- **Push-notification capture sink** — assertion grammar for "did the seller emit the completion webhook? was it RFC 9421-signed with the right key id? was the body shape correct?"
- **Intermediate-state observer** — assertion on \`working\` / \`input_required\` transitions.
- **Task cancel / resume harness** — drive a task mid-flight, cancel, assert on resulting state and cleanup.

All three depend on the TaskStore + push-notification + middleware hooks that are themselves deferred per the framework's A2A roadmap. Shipping basic \`invoke()\` now buys adopters the boilerplate elimination they need today; the lifecycle harness arrives when the framework hooks do.

## What was tested

- \`pytest tests/test_seller_a2a_client.py -v\` — 9 new tests covering success path, structured error extraction (code/message/recovery/field), unstructured-failure synthesis (unknown skill), data=None on error, structured_content always populated, public import path.
- \`ruff check src/adcp/testing/\` — clean.
- \`mypy src/adcp/testing/harness.py --ignore-missing-imports\` — clean.

### DIFF
--- a/src/adcp/testing/__init__.py
+++ b/src/adcp/testing/__init__.py
@@ -21,7 +21,12 @@ Example:
 from __future__ import annotations
 
 from adcp.testing.decisioning import build_asgi_app, build_test_client, make_request_context
-from adcp.testing.harness import AdcpErrorPayload, SellerTestClient, ToolInvokeResult
+from adcp.testing.harness import (
+    AdcpErrorPayload,
+    SellerA2AClient,
+    SellerTestClient,
+    ToolInvokeResult,
+)
 from adcp.testing.test_helpers import (
     CREATIVE_AGENT_CONFIG,
     TEST_AGENT_A2A_CONFIG,
@@ -40,6 +45,7 @@ from adcp.testing.test_helpers import (
 
 __all__ = [
     "AdcpErrorPayload",
+    "SellerA2AClient",
     "SellerTestClient",
     "ToolInvokeResult",
     "build_asgi_app",
--- a/src/adcp/testing/harness.py
+++ b/src/adcp/testing/harness.py
@@ -1,10 +1,11 @@
-"""SellerTestClient — in-process MCP harness for AdCP seller unit tests."""
+"""SellerTestClient / SellerA2AClient — in-process AdCP harnesses for unit tests."""
 
 from __future__ import annotations
 
 import asyncio
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
+from uuid import uuid4
 
 if TYPE_CHECKING:
     from adcp.decisioning import DecisioningPlatform
@@ -72,8 +73,9 @@ class SellerTestClient:
     or SSE framing. For HTTP-level tests (auth middleware, CORS, size
     limits), use :func:`adcp.testing.build_test_client` directly.
 
-    A2A transport is not yet supported (A2A is served via a separate
-    ASGI app; tracked as a follow-up on #662).
+    For A2A transport, use :class:`SellerA2AClient` — same call
+    shape (:meth:`SellerA2AClient.invoke`), different in-process
+    dispatch path (executor + event queue rather than MCP tool call).
 
     ``run_scenario()`` is not implemented — it requires bundled
     compliance scenario playbooks that are not yet available in this SDK.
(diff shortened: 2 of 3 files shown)


bokelley:
## Review

All 13 CI jobs green. The boilerplate elimination matches `SellerTestClient` so adopters with mixed MCP/A2A test suites get a single grammar — that consistency is the win.

`mergeStateStatus: BLOCKED` here just means review-required, not check failure.

### Solid

- Same `ToolInvokeResult` + `AdcpErrorPayload` return shape as the MCP client. An adopter porting an MCP assertion to A2A only changes the fixture, not the assertions. That's the right call.
- `_executor_lock` + `to_thread(self._build_executor_sync)` is correct: `create_adcp_server_from_platform` calls `asyncio.run` internally via `validate_capabilities_response_shape`, which would explode if invoked from an already-running loop. The fix is right, but **the root cause is in the framework, not this client** — see follow-up below.
- Unstructured-failure synthesis (`INTERNAL_ERROR` envelope for unknown skill / unparseable message) closes a real ergonomics gap. Without it, every adopter would have to handle structured vs. unstructured failures with two assertion patterns.
- Strict envelope validation (raise `RuntimeError` if `adcp_error.code`/`adcp_error.message` is missing) catches non-conformant servers loudly. Good defensive posture for a test client.

### Non-blocking observations

1. **`for _ in range(32):` is a magic number.** The comment says "bounded to a small number of intermediate state events" — fine for current A2A behavior, but the cap should be a constructor kwarg. If a future intermediate-state observer (one of the deferred follow-ups in the PR body) ever lands, that 32 will start mattering. Suggest `max_events: int = 32` as a constructor arg now to avoid an API break later.

2. **`EventQueueLegacy` import.** The class name advertises its own deprecation. If `a2a.server.events` has a non-legacy queue, prefer it; if not, file an upstream tracking issue. Will add this to my follow-up batch.

3. **Test gap: validation round-trip.** Constructor accepts a `validation: ValidationHookConfig` param documented as "Pass `SERVER_DEFAULT_VALIDATION` to match production behavior" — but no test asserts that an invalid payload is rejected when validation is wired. With validation defaulting to `None` (off) and not tested in either configuration, an adopter who sets `validation=SERVER_DEFAULT_VALIDATION` is trusting the executor to enforce. One test that flips the switch and asserts the rejection path would close that gap.

4. **`role="ROLE_USER"` as a string literal.** Works because protobuf accepts the enum value name as a string, but if `a2a.types.Role.USER` (or similar) is available, prefer it — the protobuf enum migration risk is real and pinning by name dodges that.

5. **Skill-name validation deferred to server.** A typo'd skill returns `INTERNAL_ERROR` synthesized envelope. That's correct (mirrors prod behavior — unknown skills fail at the executor), but the test that asserts this (`test_a2a_invoke_unknown_skill_surfaces_failure`) is the *only* coverage of the unstructured-failure synthesis path. Worth a second test for "no parseable skill at all" (DataPart present but no `skill` key) to exercise the same path with a different upstream cause.

### Deferred follow-ups already in PR body

The three #678 follow-ups (push-notification capture, intermediate-state observer, cancel/resume harness) are correctly scoped out. None of them block this PR — the basic invoke path is the most-used surface and unlocks the largest swath of adopter test rewrites.

LGTM modulo the follow-ups (which I'll file). The validation round-trip test is the only one I'd ask for in this PR specifically; the rest are issues for separate work.

Issue #720, opened 2026-05-13
BearerTokenAuthMiddleware: Authorization: Bearer should be the default; legacy header should be opt-in alias


bokelley:
## Summary

``BearerTokenAuthMiddleware`` is named after the spec-canonical RFC 6750 ``Authorization: Bearer`` header but ships with a single configurable ``header_name`` that defaults to ``"authorization"`` and accepts no aliases. Adopters with early clients sending tokens via a custom header (``x-adcp-auth``, ``X-Api-Key``, etc.) face a forced binary choice during migration: lock the middleware to the legacy header (breaks every spec-compliant new client, including ``@adcp/sdk`` ``comply()``'s api_key probe path) OR switch to ``Authorization: Bearer`` (breaks every early adopter still on the legacy header).

The right model is: **``Authorization: Bearer`` is always accepted (the spec-canonical default), legacy aliases are an opt-in additive list.** Both work simultaneously; migration is a one-way drift from legacy clients to spec-compliant ones without flag days.

## Concrete adopter pain (Prebid salesagent)

The Prebid salesagent ships ``mcp_header_name="x-adcp-auth"`` because early adopters were sending tokens that way before the protocol settled on ``Authorization: Bearer``. Every authenticated MCP call works fine for those clients. But every spec-compliant client — including the SDK's own ``security_baseline`` api_key probe, which sends ``Authorization: Bearer`` per RFC 6750 — gets 401 against the same valid token:

```bash
$ curl -i -X POST <agent>/mcp \
    -H "Authorization: Bearer <TOKEN>" \
    -H "Content-Type: application/json" -d '<jsonrpc body>'
HTTP/2 401
www-authenticate: Bearer
{"error":"unauthenticated"}

$ curl -i -X POST <agent>/mcp \
    -H "x-adcp-auth: <TOKEN>" \
    -H "Content-Type: application/json" -d '<jsonrpc body>'
HTTP/2 200
```

Net consequence: the salesagent fails the ``security_baseline/probe_api_key`` and ``assert_mechanism`` storyboard steps despite having RFC-correct auth — its auth path simply doesn't read the spec header. Filed locally as ``bokelley/salesagent#386`` with a wrapper-middleware workaround that normalizes ``Authorization: Bearer`` → ``x-adcp-auth`` before delegating to ``BearerTokenAuthMiddleware``. Every adopter mid-migration is going to write the same wrapper.

## Proposed API

[...]

Issue #714, opened 2026-05-13
IdempotencyStore.wrap should inject 'replayed: true' on cache hit (AdCP L1/security rule 4)


bokelley:
## Summary

The library's ``IdempotencyStore.wrap`` returns cached responses verbatim on cache hit. Per AdCP L1/security idempotency rule 4 — and per the library's own module-level docstring — the envelope MUST be marked ``replayed: true`` so buyer agents can suppress side effects (notifications, downstream tool calls, memory writes) on retry. Every seller using ``IdempotencyStore`` today has to subclass and reimplement ``wrap`` just to inject one boolean.

## Contradictory docs on the same behaviour

Two docstrings in the same package give opposite answers about whose responsibility this is:

``adcp/server/idempotency/store.py`` (the wrap path itself) — library does it:

> "On cache hit with matching canonical payload hash: return the cached response and mark ``replayed=True`` on the envelope."

``adcp/server/idempotency/backends.py`` (``CachedResponse`` dataclass) — seller does it:

> "Returned verbatim on replay — the seller injects ``replayed: true`` at the envelope level before sending."

Actual code: the wrap method does not inject. The seller-injection path is the live one. The store-level docstring is aspirational, the backends docstring matches reality.

## Why this is upstream's responsibility

1. **AdCP spec mandates the flag.** L1/security idempotency rule 4 makes ``replayed: true`` a wire requirement, not an optional convenience.
2. **The compliance probe asserts on it.** The ``@adcp/sdk`` ``idempotency/replay_same_payload`` storyboard verifies envelope-level ``replayed=true`` on cache hit; sellers missing it fail compliance.
3. **The injection point is in the library's wrap path.** The seller can't inject earlier (request hasn't been cached yet) or later (the wire response is already serialized) — the only natural place is exactly where ``IdempotencyStore.wrap`` returns the cached response.
4. **Every seller pays the same tax.** The salesagent fork ships a ``_ReplayMarkingStore(IdempotencyStore)`` subclass that imports four private library symbols (``_WRAPPED_FUNCTIONS``, ``_clone_response``, ``_resolve_call_args``, ``_to_dict``) just to reimplement the wrap path with the one-line injection. The class brittlely fails on any future minor bump that renames any of those symbols (we have an explicit ``ImportError`` boot-validator that raises with a pinned-version hint). Other adopters either replicate that fragility or silently fail compliance.

## Salesagent's workaround

[...]

Issue #712, opened 2026-05-13
BearerTokenAuthMiddleware._unauthenticated omits WWW-Authenticate header (RFC 6750 §3)


bokelley:
## Summary

``BearerTokenAuthMiddleware._unauthenticated()`` in ``adcp/server/auth.py:411-412`` returns a ``401`` response without a ``WWW-Authenticate`` header, violating RFC 6750 §3 / RFC 7235 §3.1. Every adopter using ``adcp.server.serve()`` with bearer-token auth on the MCP leg has this gap; compliance suites that assert on the header — including the ``security_baseline/probe_unauth`` storyboard from ``@adcp/sdk`` ``comply()`` — fail against any such deployment.

The sibling ``A2ABearerAuthMiddleware._send_unauthenticated()`` in the same file (line 1024) emits the header correctly:

```python
(b"www-authenticate", b'Bearer realm="a2a", error="invalid_token"'),
```

So the reference implementation already exists in the same module. The MCP-leg ``BaseHTTPMiddleware.dispatch()`` path just diverged.

## Current code (5.3.0)

```python
def _unauthenticated(self) -> JSONResponse:
    return JSONResponse(self._unauth_body, status_code=401)
```

Called from three sites covering the missing-token, validator-raised, and validator-returned-None paths (lines 352, 366, 369).

## Why this matters

- **RFC compliance:** Per RFC 6750 §3, every 401 from a Bearer-protected resource MUST include a ``WWW-Authenticate: Bearer ...`` challenge. RFC-compliant clients (browsers, many HTTP libraries) don't surface auth failures to the user without it.
- **AdCP compliance:** The ``security_baseline`` universal storyboard's ``probe_unauth`` step asserts on header presence at the buyer-protocol layer. The storyboard accepts any 401-with-header response and fails 401-without-header.
- **Buyer-agent self-correction:** LLM-driven buyer agents that walk ``WWW-Authenticate`` to choose the auth scheme silently drop the response when the header is absent and retry indefinitely against the same endpoint.

## Reproduction

A 7.1.0 ``comply()`` run against ``https://wonderstruck.sales-agent.scope3.com/mcp`` (adcp 5.3.0 deployed) surfaces:

```
storyboard: security_baseline
step:       probe_unauth
task:       $test_kit.auth.probe_task
expected:   401 with WWW-Authenticate: Bearer realm="..."
actual:     401 with body {"error":"unauthenticated"}, no WWW-Authenticate
[...]

PR #390, opened 2026-05-03
docs(examples): Alembic migration scaffold for v3 reference seller

bokelley:
Refs #382

Adds a complete Alembic migration scaffold to `examples/v3_reference_seller/` so production adopters have a real schema-evolution story instead of `create_all`.

## What changed

**New files:**
- `alembic.ini` — Alembic config; `DATABASE_URL` read from env (never hardcoded)
- `alembic/env.py` — async engine setup (asyncpg); explicitly imports both `src.models` and `src.audit` so all 5 tables appear in autogenerate
- `alembic/script.py.mako` — standard migration template
- `alembic/versions/0001_initial_schema.py` — hand-written initial migration covering all 5 tables (tenants, buyer_agents, accounts, media_buys, audit_events) with indexes, FKs, and CHECK constraints
- `migrate.py` — standalone script that runs `alembic upgrade head`; redacts DB password before logging
- `tests/test_migrations.py` — two `@pytest.mark.integration` tests (skipped without `DATABASE_URL`); use SQLAlchemy async engine for all inspection (no psycopg2 dependency)

**Modified:**
- `README.md` — removed Alembic from "What's NOT wired" bullet; added `## Migrations` section with install, apply, generate, rollback, and test-run instructions

## Deliberate deviation from acceptance criterion #3

The issue asks the boot path to call `alembic upgrade head` instead of `create_all`. DX review flagged this as a blocker: making Alembic a hard runtime dependency at boot produces opaque errors for first-time adopters and silently breaks any checkout without Alembic installed (the example has no `requirements.txt`). Instead:
- `src/app.py` boot path is **unchanged** (`create_all` — fast iteration default)
- `migrate.py` provides the production migration path as an explicit step
- The README `## Migrations` section explains the two-path model clearly, with the `create_all` warning prominent

## What was tested

- `pytest examples/v3_reference_seller/tests/test_smoke.py -v` — 7/7 passed
- `ruff check examples/v3_reference_seller/` — clean
- `mypy` not run (no SDK type changes; examples/ is not in mypy scope)
- Integration tests (`test_migrations.py`) require a live Postgres instance; skipped in this environment

## Pre-PR review

- code-reviewer: approved — blockers fixed (type narrowing in env.py, psycopg2 assumption in tests, downgrade test ordering, credential leak in migrate.py)
- docs-expert: approved — README section is accurate, operational, and audience-appropriate; `integration` marker pre-declared in pyproject.toml (no fork friction)

[...]

PR #732, opened 2026-05-14
feat(decisioning): PgProposalStore + framework package derivation (#727)

### LINKED ISSUE #727: ProposalStore: ship PgProposalStore + framework-level package derivation from allocations
## Summary

Two adopter-side pieces every production seller wires today that belong upstream. Same upstream-vs-local-workaround pattern as #712 (WWW-Authenticate), #714 (replayed: true), and #720 (multi-header auth) — proven productive this week (all three shipped in 5.4.0 and adopter code shrunk).

## (1) ``PgProposalStore`` — concrete durable backing

Upstream ships ``adcp.decisioning.InMemoryProposalStore`` (non-durable reference) + the ``ProposalStore`` Protocol. Every production adopter writes the same Postgres-backed implementation:

* The same nine methods (``put_draft`` / ``get`` / ``commit`` / ``try_reserve_consumption`` / ``finalize_consumption`` / ``release_consumption`` / ``mark_consumed`` / ``discard`` / ``get_by_media_buy_id``).
* The same ``SELECT … FOR UPDATE`` row-lock pattern for the CAS in ``try_reserve_consumption`` (Protocol contract).
* The same composite primary key shape: ``(tenant_id, proposal_id)`` for multi-tenant scoping; partial unique on ``(tenant_id, media_buy_id) WHERE media_buy_id IS NOT NULL`` for ``get_by_media_buy_id``.
* The same state-machine guards (``commit`` from non-DRAFT raises INTERNAL_ERROR, ``try_reserve_consumption`` from non-COMMITTED raises PROPOSAL_NOT_COMMITTED, etc.).

The Prebid salesagent fork shipped this as ``SalesAgentProposalStore`` in [bokelley/salesagent#390](https://github.com/bokelley/salesagent/pull/390) (~430 LOC core + tests). The implementation is mostly mechanical — same shape as ``IdempotencyStore.PgBackend`` adopters get for free.

### Proposed API

```python
from adcp.decisioning.proposal_store import PgProposalStore
from psycopg_pool import AsyncConnectionPool

pool = AsyncConnectionPool(DATABASE_URL, ...)
store = PgProposalStore(
    pool=pool,
    # Adopter chooses the table name to avoid collisions with existing
    # ``proposal_drafts`` / ``proposals`` tables. Defaults to the canonical
    # name; explicit prefix lets adopters with one Postgres serving multiple
    # adcp instances keep them scoped.
    table_name="adcp_proposal_drafts",
)
PgProposalStore.MIGRATION  # alembic-compatible upgrade/downgrade strings
```

Same shape as ``IdempotencyStore.PgBackend`` ships today.

## (2) Framework-level package derivation from proposal allocations

The framework already:

* Reserves the proposal via ``try_reserve_consumption`` (proposal_dispatch.py:585-588)
* Hydrates ``ctx.recipes`` (line 607)
* Documents that "the spec allows the seller to derive packages from the proposal's allocations" (line 590-593)

But then it punts: the seller's ``platform.create_media_buy`` is handed an empty ``req.packages`` and is expected to do the derivation. Every adopter writes essentially the same percentage math:

```python
for allocation in proposal_payload["allocations"]:
    pkg_budget = total_budget.amount * (allocation["allocation_percentage"] / 100.0)
    packages.append(PackageRequest(
        product_id=allocation["product_id"],
        budget=pkg_budget,
        pricing_option_id=allocation.get("pricing_option_id"),
    ))
```

The salesagent had to ship this seller-side just to pass the ``proposal_finalize/create_media_buy`` storyboard. So will every other seller declaring ``capabilities.refine=True`` or wiring a ProposalStore.

### Proposed API — framework-level auto-injection

In ``proposal_dispatch.maybe_hydrate_recipes_for_create_media_buy``, after reserving the proposal: if ``req.packages`` is empty AND the proposal has ``allocations[]``, mutate ``req.packages`` to the derived list before dispatching to ``platform.create_media_buy``.

The seller's adapter sees a normal ``create_media_buy`` with populated packages — never needs to know it came from a proposal. Existing ``packages``-based paths (no ``proposal_id``) unchanged.

### Override hook for non-default strategies

Some sellers will want custom derivation logic — auction pricing options need ``bid_price``, multi-currency proposals need explicit currency, capability-overlap filtering may shape the package count. Expose a method on ``ProposalManager``:

```python
class MyProposalManager:
    def derive_packages(
        self,
        proposal_payload: dict,
        total_budget: TotalBudget,
        recipes: Mapping[str, Recipe],
    ) -> list[PackageRequest]:
        \"\"\"Custom derivation. Framework calls this when proposal_id is set
        and req.packages is empty. Default implementation (when this method
        is absent) does even-percentage distribution per allocation.\"\"\"
```

Adopters with default behavior (even percentage split) leave the method off; the framework falls through to its built-in even-split derivation. ``SalesAgentProposalManager`` and most other minimal adopters delete their custom code; sellers with auction / multi-currency / complex inventory shaping override.

### Standalone helper for sellers that need it

Even if (2) ships, expose ``adcp.decisioning.derive_packages_from_proposal(...)`` as a public helper so adopters with workflows outside ``proposal_dispatch``'s auto-injection path can still call it. Mirror the shape ``IdempotencyStore.wrap`` exposes — auto-applied via decorator, also callable directly.

## Why now

Five days, four PRs (#712, #714, #720, plus the implicit #722/#723 wave), every one collapsed adopter code by 50-500 LOC. The pattern is reliable: defects and patterns in adopter ``decisioning/`` wiring are good candidates for the library. This issue closes two more.

## Related

- bokelley/salesagent#387 — adopter-side ProposalStore wiring tracker (closed when #390 + the upcoming derivation PR land)
- bokelley/salesagent#390 — ``SalesAgentProposalStore`` PR (collapses to ~10 LOC once PgProposalStore ships upstream)
- adcp 5.4.0 PRs #722/#724 / #723/#725 — adopter-API surface (factory kwargs + auto_commit_on_put_draft); this issue is the natural next step (concrete backings + auto-derivation)

bokelley:
## Summary

Closes #727. Two patterns moved upstream from the salesagent fork ([bokelley/salesagent#390](https://github.com/bokelley/salesagent/pull/390)):

### 1. `PgProposalStore` — durable `ProposalStore` Protocol implementation
- Mirrors the existing `PgTaskRegistry` / `IdempotencyStore.PgBackend` patterns (pool injection, `is_durable=True`, `create_schema()` idempotent bootstrap, ASCII-safe identifier guards).
- `try_reserve_consumption` CAS uses `SELECT ... FOR UPDATE` inside a transaction; `commit()` does the same to close a SELECT-then-UPDATE race that the security/python reviewers caught mid-PR.
- `PgProposalStore.migration_sql(table_name=...)` classmethod for Alembic/dbmate adopters (honours custom table names — previous hardcoded `MIGRATION` constant was a footgun).
- `recipe_decoder` callable for adopters with typed `Recipe` subclasses; default decoder fails loudly with a pointer to `recipe_decoder=` when a typed payload is detected.

### 2. Framework package derivation from proposal allocations
- Opt-in via `ProposalCapabilities.derive_packages_from_allocations=True` OR by implementing `ProposalManager.derive_packages(...)` — default off preserves pre-#727 semantics for existing adopters who read `ctx.recipes` directly.
- When enabled and the buyer omits `packages[]` on `create_media_buy(proposal_id=..., total_budget=...)`, framework mutates `req.packages` from the proposal's `allocations[]` via percentage math. `start_time` / `end_time` propagate per spec ("per-flight scheduling within a proposal").
- Standalone `adcp.decisioning.derive_packages_from_proposal(...)` helper for off-path workflows (admin tools, draft-preview UIs).
- Auto-picks the product's single `pricing_options[]` entry at proposal-persist time when an allocation omits `pricing_option_id` (unambiguous case); multi-option products require explicit allocation-level selection or a `derive_packages` override.

### Security
Closes the cross-tenant write surface on `commit()` / `mark_consumed()` / `discard()` by adding required `expected_account_id` to the Protocol. Without it, the PG store's `(account_id, proposal_id)` PK could be hit on the wrong tenant's row via colliding `proposal_id`s. All three methods now scope at the SQL level.

### Reviewer rollcall
Reviewed by `code-reviewer`, `security-reviewer`, `ad-tech-protocol-expert`, `adtech-product-expert`, `python-expert`, `dx-expert`. All Must-Fix items addressed; a few Should-Fix items deferred to follow-ups (`is_durable` boot-gate parity for ProposalStore; `Recipe.from_kind_payload` registry to eliminate `recipe_decoder` from the hello-world path).

## Test plan
- [x] Full unit suite passes (4328 passed locally; 1 unrelated flaky timing test on `test_pg_idempotency_backend`)
- [x] Mypy clean across 805 source files
- [x] Ruff clean across changed files
- [x] New unit tests: `tests/test_decisioning_pg_proposal_store.py` (structural, no DB)
- [x] New conformance tests: `tests/conformance/decisioning/test_pg_proposal_store.py` (real-PG; skips without `ADCP_PG_TEST_URL`) — includes the cross-tenant rejection and the `SELECT ... FOR UPDATE` concurrency race
- [x] New unit tests: `tests/test_derive_packages.py` (pure-function math + integration with `maybe_hydrate_recipes_for_create_media_buy`)
- [x] Existing `tests/test_proposal_lifecycle_e2e.py` (21 tests) passes with the opt-in flag flipped on the example seller's manager

### DIFF
--- a/examples/sales_proposal_mode_seller/src/proposal_manager.py
+++ b/examples/sales_proposal_mode_seller/src/proposal_manager.py
@@ -121,8 +121,19 @@ def _draft_proposal_payload(allocations: dict[str, float] | None = None) -> dict
     Required fields: ``proposal_id``, ``name``, ``allocations``. Each
     allocation entry carries ``product_id`` + ``allocation_percentage``
     (0-100). Percentages must sum to 100.
+
+    The ``pricing_option_id`` on each allocation feeds the framework's
+    package derivation (per #727) — when the buyer calls
+    ``create_media_buy(proposal_id=..., total_budget=...)`` with no
+    explicit ``packages[]``, the framework derives one package per
+    allocation using ``pricing_option_id``.
     """
     alloc = allocations or {"ctv-premium-q2": 60.0, "display-run-q2": 40.0}
+    # Map each product to the canonical pricing_option_id from _PRODUCTS.
+    pricing_option_ids = {
+        "ctv-premium-q2": "po-ctv-cpm",
+        "display-run-q2": "po-display-cpm",
+    }
     return {
         "proposal_id": PROPOSAL_ID,
         "name": "Balanced Reach Q2 — CTV-led",
@@ -132,6 +143,7 @@ def _draft_proposal_payload(allocations: dict[str, float] | None = None) -> dict
             {
                 "product_id": pid,
                 "allocation_percentage": pct,
+                "pricing_option_id": pricing_option_ids.get(pid, "po-default"),
                 "rationale": f"Indicative {pct:.0f}% allocation to {pid}.",
             }
             for pid, pct in alloc.items()
@@ -154,6 +166,9 @@ class ProposalModeProposalManager:
         # 60s grace absorbs clock skew between the seller's
         # expires_at and the buyer's create_media_buy call.
         expires_at_grace_seconds=60,
+        # #727: let the framework derive packages from the proposal's
+        # allocations when the buyer omits packages[] on create_media_buy.
+        derive_packages_from_allocations=True,
     )
 
     async def get_products(
--- a/src/adcp/decisioning/__init__.py
+++ b/src/adcp/decisioning/__init__.py
@@ -84,6 +84,7 @@ from adcp.decisioning.context import (
     AuthInfo,
     RequestContext,
 )
+from adcp.decisioning.derive_packages import derive_packages_from_proposal
 from adcp.decisioning.dispatch import validate_platform
 from adcp.decisioning.errors import (
     AccountNotFoundError,
@@ -260,7 +261,11 @@ from adcp.decisioning.validate_capabilities import (
 # ``Pg*`` convention shared with PgReplayStore / PgBuyerAgentRegistry /
 # PgWebhookDeliverySupervisor.
 try:
-    from adcp.decisioning.pg import PgTaskRegistry, PostgresTaskRegistry  # noqa: F401
+    from adcp.decisioning.pg import (  # noqa: F401
+        PgProposalStore,
+        PgTaskRegistry,
+        PostgresTaskRegistry,
+    )
 except ImportError:  # pragma: no cover — exercised by the [pg] extra tests
     from typing import ClassVar as _ClassVar
 
@@ -282,6 +287,22 @@ except ImportError:  # pragma: no cover — exercised by the [pg] extra tests
 
     # Deprecated alias preserved through 4.4.x.
     PostgresTaskRegistry: type[PgTaskRegistry] = PgTaskRegistry  # type: ignore[no-redef]
+
+    class PgProposalStore:  # type: ignore[no-redef]
+        """Stub raised when ``adcp[pg]`` isn't installed.
+
+        Attempting to instantiate raises :class:`ImportError` with the
+        install-hint text from :mod:`adcp.decisioning.pg.proposal_store`.
+        """
+
+        is_durable: _ClassVar[bool] = True
+
+        def __init__(self, *args: object, **kwargs: object) -> None:
+            raise ImportError(
+                "PgProposalStore requires psycopg3 and psycopg-pool. "
+                "Install the 'pg' extra: `pip install 'adcp[pg]'` "
+                "(Poetry: `poetry add 'adcp[pg]'`)."
+            )
 
 
 __all__ = [
@@ -338,6 +359,7 @@ __all__ = [
     "NoAuth",
     "OAuthCredential",
     "PermissionDeniedError",
+    "PgProposalStore",
     "PgTaskRegistry",
     "LazyPlatformRouter",
     "PlatformRouter",
@@ -415,6 +437,7 @@ __all__ = [
     "create_tenant_store",
     "create_translation_map",
     "create_upstream_http_client",
+    "derive_packages_from_proposal",
     "require_account_match",
     "require_advertiser_match",
     "require_org_scope",
--- a/src/adcp/decisioning/derive_packages.py
+++ b/src/adcp/decisioning/derive_packages.py
@@ -1,0 +1,231 @@
+"""Framework-level package derivation from proposal allocations.
+
+When a buyer calls ``create_media_buy(proposal_id=..., total_budget=...)``
+without supplying ``packages[]``, the spec lets the seller derive
+packages from the committed proposal's ``allocations[]`` array via
+percentage math. Every production adopter writes essentially the same
+loop:
+
+.. code-block:: python
+
+    for allocation in proposal_payload["allocations"]:
+        pkg_budget = total_budget.amount * (allocation["allocation_percentage"] / 100.0)
+        packages.append(PackageRequest(
+            product_id=allocation["product_id"],
+            budget=pkg_budget,
+            pricing_option_id=allocation["pricing_option_id"],
+        ))
+
+The framework auto-invokes :func:`derive_packages_from_proposal` from
+:func:`adcp.decisioning.proposal_dispatch.maybe_hydrate_recipes_for_create_media_buy`
+when ``req.packages`` is empty AND the proposal has ``allocations[]``,
+so seller adapters see a normal ``create_media_buy`` with populated
+packages — they never need to write the loop.
+
+Adopters with custom derivation logic (auction pricing options needing
+``bid_price``, multi-currency proposals, capability-overlap filtering)
+override the behavior by implementing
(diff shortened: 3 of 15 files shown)


bokelley:
Filed #727; happy to see both halves landing together. **Salesagent will be the reference downstream adopter.**

Mapping our two staged PRs onto this:

| Salesagent | Upstream in #732 | Action on tag |
|---|---|---|
| [PR #390 — `SalesAgentProposalStore`](https://github.com/bokelley/salesagent/pull/390) (deployed) | `PgProposalStore` | Swap implementation, delete custom file (~280 LOC) |
| [PR #406 — `derive_packages_from_proposal` + delegate hook](https://github.com/bokelley/salesagent/pull/406) (review-ready) | `derive_packages_from_allocations=True` + framework auto-injection | Flip the capability flag, delete `core/proposal/derivation.py` (~170 LOC) + delete the delegate's proposal_id branch (~40 LOC) |

Net delete on adoption: ~490 LOC. Same shape as the #712 / #714 / #720 wave — file upstream, ship local for compliance, delete local when tag.

**One thing worth a checkpoint before merge:**

Salesagent mints `account_id` as compound `f\"{tenant_id}:{ref}\"` (multi-tenant subdomain deploys; the `ref` is the principal identity, `tenant_id` is the FK target for our `proposals.tenant_id` column). The framework passes `ctx.account.id` (the compound string) through to `ProposalStore.put_draft` / `get` / `commit` / etc. unchanged today.

For our adoption to be a clean drop-in, `PgProposalStore` needs to:
1. **Treat `account_id` as an opaque string** for equality filtering — no parsing, no joins on it. The description implies this is true (`(account_id, proposal_id)` PK, scoped equality filters in WHERE clauses).
2. **Not require a 1:1 mapping** between `account_id` and any other Protocol-visible identity. We need to derive `tenant_id` (FK target) from `account_id` ourselves; we'd do that in a subclass / extension hook, not by patching the Protocol.

If both hold, our compound format works without any upstream concession. Just flagging for the python-expert pass — easy to miss because the storyboard examples use opaque ULIDs.

**On `recipe_decoder=`**: confirming this sees the raw JSONB dict (not a pre-decoded `Recipe` instance) so default decoders can handle the `dict[str, Recipe]` shape `SalesAgentProposalManager` persists. If yes, our adoption is `recipe_decoder=None` (default).

Will adopt within a sprint of the tag landing. Happy to ride a pre-release if useful for catching the compound-`account_id` integration before tag.

bokelley:
Confirmed on both checkpoints — `PgProposalStore` is shape-compatible with the salesagent compound `account_id`.

### `account_id` is opaque

Audited `src/adcp/decisioning/pg/proposal_store.py` end-to-end. Every reference to `account_id` is either:
- Part of a SQL identifier (column name, PK, index)
- A `%s`-parameterized equality predicate (`WHERE account_id = %s`)
- An unmodified passthrough into the bind tuple

The only Python-side operation on the string is `if expected_account_id is None` (the optional get-without-tenancy gate). Zero `split` / `partition` / `startswith` / regex / slicing — `grep -nE 'account_id\.(split|partition|replace|startswith|endswith|find|index|count)\|account_id\['` returns empty.

Your compound `f"{tenant_id}:{ref}"` round-trips byte-for-byte. The salesagent's derivation of `tenant_id` from `account_id` is purely your concern (in a subclass / extension hook) — the framework never inspects the format.

### `recipe_decoder` sees the raw JSONB dict

`_decode_recipes` calls `decoder(payload)` where `payload: Mapping[str, Any]` is the parsed JSONB value for a single `product_id` — i.e. the dict produced by `Recipe.model_dump(mode="json")` at persist time. The decoder gets a plain dict, NOT a pre-decoded Recipe instance.

For `SalesAgentProposalManager`'s `dict[str, Recipe]` shape: encoder calls `recipe.model_dump(mode="json")` per entry; decoder receives that dict back. The default decoder (`Recipe.model_validate`) handles the base shape, so if your recipes don't declare typed subclass fields, `recipe_decoder=None` (the default) is correct. If you add a typed subclass later, the default decoder fails loudly with a pointer to `recipe_decoder=` — no silent corruption.

### Salesagent adoption flag

`ProposalCapabilities.derive_packages_from_allocations=True` activates the framework's auto-injection. Either that OR implementing `ProposalManager.derive_packages(...)` triggers derivation; default `False` preserves pre-#727 semantics for adopters who haven't opted in. So PR #406's delete is mechanical: flip the cap flag, delete `core/proposal/derivation.py`, delete the delegate's `proposal_id` branch.

Happy to cut a pre-release tag once this lands so you can pin against it before the official release.

Issue #375, opened 2026-05-03
Tier 2 commercial-identity error codes don't conform to AdCP spec — cross-tenant oracle risk


bokelley:
## Summary

The Python SDK raises four non-spec error codes in `src/adcp/decisioning/handler.py` and `src/adcp/decisioning/registry.py` for the v3 buyer-agent gate. The codes the spec rejects are emitted today as load-bearing public wire errors:

- `AGENT_SUSPENDED`
- `AGENT_BLOCKED`
- `REQUEST_AUTH_UNRECOGNIZED_AGENT`
- `INVALID_BILLING_MODEL`

## Why it's wrong

The AdCP spec at `static/schemas/source/error-details/agent-permission-denied.json` (line 10) explicitly considers and rejects separate codes for these scenarios on cross-tenant onboarding-oracle grounds:

> If a future major version elevates `suspended` / `blocked` to dedicated error codes (e.g., `AGENT_SUSPENDED` / `AGENT_BLOCKED`), those codes will carry their own details shapes and SHOULD NOT carry an explicit `scope` field — the code itself is the discriminator on that path.

And, more importantly, the schema description and `error-handling.mdx` (Per-Agent Authorization Gate) require **identical observable channels** — code, message, headers, latency, side effects — between the unrecognized-agent path and the recognized-but-denied path. Distinct codes per status leak which `agent_url`s are onboarded with which sellers, enabling cross-tenant enumeration of commercial relationships.

### Code surface

- `AGENT_SUSPENDED` → `PERMISSION_DENIED` with `details.scope=\"agent\"`, `details.status=\"suspended\"`. Recovery is currently `transient`; the spec treats suspension as terminal-pending-onboarding (the agent MUST surface to a human; auto-retry is not a recovery path).
- `AGENT_BLOCKED` → `PERMISSION_DENIED` with `details.scope=\"agent\"`, `details.status=\"blocked\"`, recovery terminal.
- `REQUEST_AUTH_UNRECOGNIZED_AGENT` → `PERMISSION_DENIED` with `scope` **omitted** (must be indistinguishable on the wire from recognized-but-denied; the schema's omit-on-unestablished-identity rule).
[...]

PR #735, opened 2026-05-14
fix(server): Tier 2 permission-denied parity contract (#392)

### LINKED ISSUE #392: Tier 2 commercial-identity gate: latency / headers / side-effects parity contract
## Summary

Follow-up from #375 (PR for the wire-code rename portion). This issue tracks the **parity-contract refactor** that was deferred from that PR per the user's stop-rule (\"if the parity contract turns out to be more complex than fits in a single PR, file a separate issue for the parity-contract follow-up\").

## Why

The cross-tenant onboarding-oracle clamp in the AdCP spec requires the unrecognized-agent path and the recognized-but-denied path to be observably **indistinguishable** to an external attacker — across:

- HTTP status code (already aligned via shared `PERMISSION_DENIED` envelope).
- Response headers (Content-Type, Content-Length within negligible tolerance).
- Side effects (audit emission, metric increment with the same operation label).
- Latency (cannot shortcut the unrecognized path; needs a deliberate budget).
- Observability (logs, APM spans, third-party error telemetry).

#375's rename portion closed the code mismatch but left the eager-raise structural difference — see the docstring on `_resolve_buyer_agent` (the \"Note on parity\" paragraph) and the test-file docstring on `tests/test_tier2_spec_conformance.py`.

## Required design

Per the issue body of #375:

> Approach: introduce a `PermissionDeniedReason` internal enum/struct with `(scope, status)` fields. The dispatch layer raises a single `PermissionDeniedError` carrying a reason; the wire-level translator emits the spec-correct `PERMISSION_DENIED` envelope with `details` populated only when scope is set. The unrecognized path still does the same audit-emission + metric increment as the denied path. Latency parity may require an explicit small delay budget — document the tradeoff.

Concretely:

1. **Single emit point.** Move all four denial paths in `_resolve_buyer_agent` (and the `validate_billing_for_agent` path) through one common projection — an internal `PermissionDeniedError` wrapping a `PermissionDeniedReason` struct, projected to the wire envelope by a single translator. Today four call sites build the envelope inline.
2. **Latency budget.** The unrecognized path needs to do equivalent resolution work to the recognized-but-denied path before raising. The simplest design is an explicit small delay (50–100 ms?) on every denial path so the variance between branches is dominated by the budget rather than the code path. Document the budget tradeoff (extra latency on every denial vs. closing the side channel).
3. **Audit / metric parity.** Today the recognized branches log `details.agent_url` and `details.status`; the unrecognized branch logs nothing. Need to converge on a uniform log shape — same operation label, same fields (with the discriminator either missing or hashed-then-truncated to prevent log scraping from becoming the side channel).
4. **Header parity.** A2A and MCP transports both go through the framework's wire serializer. Verify Content-Length is within tolerance (the message strings are already identical post-#375; the variance is `details` size). Consider padding the unrecognized envelope to match a fixed byte count.

## Tests

- Latency parity test: assert p99 difference between unrecognized-agent and recognized-but-suspended paths is < 5 ms over N requests (with the explicit budget set high enough that branch-comparison variance is negligible).
- Header parity test: same Content-Type, Content-Length within ±N bytes (where N covers the `details` shape variance).
- Side-effects parity test: same audit-sink writes, same metric increments with the same operation label.
- Status code parity test: same HTTP status across all four denial paths.

## Refs

- #375 (rename portion shipped here).
- Spec source (when available): `static/schemas/source/error-details/agent-permission-denied.json` line 10 — the elevation note.
- Spec uniform-response MUST list: `error-handling.mdx` (Per-Agent Authorization Gate).

bokelley:
## Summary

Closes #392. Closes the structural / latency / header / observability gaps in the cross-tenant onboarding-oracle clamp left over from #375.

- Single internal `PermissionDeniedError` + translator (was 4 inline envelope sites in `_resolve_buyer_agent`)
- Explicit deadline-relative latency budget on every denial path (env-tunable via `ADCP_PERMISSION_DENIED_BUDGET_MS`, default 50 ms)
- Audit/metric parity: uniform operation label `buyer_agent_registry.permission_denied`, uniform `details` key set, `agent_url` hashed-truncated via `sha256(...)[:12]` so log-scraping cannot rebuild the side channel
- Header parity tradeoff documented in the module docstring — `Content-Length` variance is bounded by `details` payload size; the spec's omit-on-unestablished-identity rule prevents padding the unrecognized envelope, and the intra-recognized variance (`agent_url` is buyer-controlled) already exceeds the recognized-vs-unrecognized delta

## Security review

Self-checks performed against the design's review points:

1. **Field-presence side channel in audit emission?** The audit row's `details` key set is identical on every denial path — `outcome`, `reason_scope`, `reason_status`, `agent_url_hash` are all present; values are `None` on the unrecognized branch. Verified by `test_audit_emit_uniform_details_key_set` and `test_audit_emit_unrecognized_carries_none_hash_not_missing_key`. On the wire, `details` is omitted entirely on unrecognized (the empty dict on `AdcpError` is dropped by `to_wire()`), preserving the existing spec-conformance behavior.

2. **Ordering leak between audit emit and budget sleep?** Identical ordering on every branch, structurally enforced by routing all branches through `raise_permission_denied`: capture deadline → emit audit → sleep until deadline → raise. The deadline-relative sleep (rather than fixed-duration) absorbs audit-sink variance so total wall-clock is dominated by the budget regardless of branch.

3. **Padding content leak?** No padding applied — the spec's omit-on-unestablished-identity rule prevents padding the unrecognized envelope. The tradeoff is documented in the module docstring with the justification that intra-recognized `agent_url` length variance already exceeds the recognized-vs-unrecognized delta.

What I'd want a security reviewer to verify next:

- **Audit-sink wiring downstream.** `PlatformHandler` now accepts `permission_denied_audit_sink` — verify that the v3 reference seller / SalesAgent wires this to the same sink as the registry-cache layer so SecOps sees registry events and denial events in a single stream.
- **Budget tuning under load.** The 50 ms default assumes audit-sink p99 < 30 ms. Real-world DB-backed sinks can spike higher; production rollouts should establish an SLO on the sink and document the budget setting alongside it.
- **Latency-parity test sensitivity in CI.** The latency-parity test isolates `raise_permission_denied` from `PlatformHandler` setup overhead so the assertion targets the parity property rather than handler jitter. Verify the test stays sub-5ms on the CI runner profile; if CI moves to noisier hardware, the budget in that test can be raised (the production setting is independent).

## Test plan

- [x] Latency parity p99 diff < 5 ms (`tests/test_tier2_parity_contract.py::test_latency_parity_p99_difference_under_budget`)
- [x] Header parity within ±200 bytes `Content-Length` (`test_header_content_length_within_tolerance`); two unrecognized envelopes byte-identical (`test_unrecognized_envelopes_are_byte_equivalent`)
- [x] Side-effects parity: audit operation label, key set, hash-truncated discriminator (`test_audit_emit_same_operation_label_across_paths`, `test_audit_emit_uniform_details_key_set`, `test_audit_emit_agent_url_is_hashed_not_plaintext`, `test_audit_emit_unrecognized_carries_none_hash_not_missing_key`)
- [x] Status code parity across 4 denial paths (`test_status_code_parity_across_four_paths`)
- [x] Audit-sink failure isolation: raising sink doesn't bypass the gate (`test_raising_sink_does_not_propagate`)
- [x] Env-var configuration: tunable, malformed-value fallback, default (`test_budget_env_var_*`)
- [x] Existing `tests/test_tier2_spec_conformance.py` still passes (12 tests green)
- [x] Full unit suite green: 4814 passed, 24 skipped, 1 xfailed
- [x] `ruff check src/` + `mypy src/adcp/` clean

### DIFF
--- a/src/adcp/decisioning/handler.py
+++ b/src/adcp/decisioning/handler.py
@@ -186,6 +186,7 @@ from adcp.types import (
 if TYPE_CHECKING:
     from concurrent.futures import ThreadPoolExecutor
 
+    from adcp.audit_sink import AuditSink
     from adcp.decisioning.platform import DecisioningPlatform
     from adcp.decisioning.property_list import PropertyListFetcher
     from adcp.decisioning.registry import BuyerAgent, BuyerAgentRegistry
@@ -441,6 +442,8 @@ SPECIALISM_TO_PROTOCOLS: dict[str, frozenset[str]] = {
 async def _resolve_buyer_agent(
     registry: BuyerAgentRegistry,
     auth_info: AuthInfo | None,
+    *,
+    audit_sink: AuditSink | None = None,
 ) -> BuyerAgent:
     """Resolve a :class:`BuyerAgent` from a wired registry.
 
@@ -477,13 +480,40 @@ async def _resolve_buyer_agent(
       ``details`` OMITTED — scope MUST NOT be set on the unestablished-
       identity path (omit-on-unestablished-identity rule).
 
-    Note on parity: the *latency / headers / side-effects* parity
-    contract between the recognized and unrecognized paths is tracked
-    as a follow-up — the eager-raise pattern below still completes the
-    unrecognized path on a different code path than the recognized
-    one. Renaming closes the wire-code mismatch; folding all four
-    paths through a common emit point with deliberate latency padding
-    and identical audit/metric side-effects is the next step.
+    Parity contract
+    ---------------
+
+    All four denial paths funnel through
+    :func:`adcp.decisioning.permission_denied.raise_permission_denied`,
+    which:
+
+    * Emits a single :class:`~adcp.audit_sink.AuditEvent` with the
+      uniform operation label ``buyer_agent_registry.permission_denied``
+      and the same key set in ``details`` on every path. The
+      ``agent_url`` is hashed-truncated before it reaches the sink so
+      log-scraping cannot reconstruct the side channel from internal
+      telemetry.
+    * Sleeps until a shared deadline
+      (``perf_counter() + ADCP_PERMISSION_DENIED_BUDGET_MS / 1000``,
+      default 50 ms). Deadline-relative rather than fixed-duration so
+      audit-sink latency variance is absorbed into the budget rather
+      than extending past it.
+    * Raises via a single translator
+      (:func:`~adcp.decisioning.permission_denied.translate_to_adcp_error`)
+      so the wire envelope shape is identical across the four
+      branches modulo the ``details`` payload (omitted on
+      unrecognized; populated with ``{scope, status, agent_url}`` on
+      recognized).
+
+    See the module docstring of :mod:`adcp.decisioning.permission_denied`
+    for the latency-budget and header-parity tradeoffs.
+
+    :param audit_sink: Optional :class:`~adcp.audit_sink.AuditSink` for
+        denial events. When ``None`` (default), denial events log at
+        ``DEBUG`` instead. Wired by ``PlatformHandler`` from the
+        adopter-supplied sink; unset paths (test fixtures, sinkless
+        deployments) still get the latency-budget and translator
+        parity guarantees.
 
     :raises AdcpError: ``PERMISSION_DENIED`` (all four denial paths).
         Recovery is ``correctable`` per the spec's ``enumMetadata``
@@ -494,11 +524,18 @@ async def _resolve_buyer_agent(
         the signal callers surface to a human operator rather than
         loop on the request.
     """
+    from adcp.decisioning.permission_denied import (
+        PermissionDeniedError,
+        PermissionDeniedReason,
+        raise_permission_denied,
+        translate_to_adcp_error,
+    )
     from adcp.decisioning.registry import (
         ApiKeyCredential,
         HttpSigCredential,
         OAuthCredential,
     )
+    from adcp.decisioning.registry_cache import _current_tenant_id
     from adcp.decisioning.types import AdcpError
 
     credential = auth_info.credential if auth_info is not None else None
@@ -514,6 +551,10 @@ async def _resolve_buyer_agent(
             # with INTERNAL_ERROR rather than silently passing the
             # request through (which would skip the registry gate
             # entirely and leak the upgrade footgun into production).
+            # NOT routed through the PERMISSION_DENIED parity gate:
+            # an INTERNAL_ERROR is an adopter-config bug, not a
+            # commercial-identity rejection, and adopters need a
+            # distinct wire signal to fix it.
             raise AdcpError(
                 "INTERNAL_ERROR",
                 message=(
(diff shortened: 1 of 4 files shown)


bokelley:
Holding for human security review before merge — this is a 1,100-line auth refactor with explicit timing-side-channel + audit/header parity concerns. The implementing agent self-flagged that an independent security-reviewer pass is the right next step.

**Not blocking the 5.5.0 release.** Will ship in a subsequent release after review.

bokelley:
**Closing as superseded by #748.**

Quick recap of why this PR's wire-level parity premise no longer holds, and where the residual value moved:

## Why this PR is superseded

PR #748 (merged 2026-05-20) migrated the recognized-but-denied paths to dedicated wire codes (`AGENT_SUSPENDED` / `AGENT_BLOCKED`) with `recovery="terminal"`. That is the spec-aligned strategy adopted by AdCP 3.1.0-beta.1 (upstream PR `adcontextprotocol/adcp#3906`).

This PR was built against the pre-#748 world where all four denial branches funneled through `PERMISSION_DENIED` with different `details` payloads. The whole "four branches funnel through a single translator" structural contract no longer applies — the recognized-but-denied branches now intentionally diverge on the wire via dedicated codes, and that's the *correct* design per the spec.

Per the protocol-expert review of #748: *"recognized-but-denied path no longer carries `scope='agent'`, so it's now indistinguishable from any vanilla `PERMISSION_DENIED`, not just from the unrecognized path."* The very wire-level oracle this PR was closing was closed differently — and better — by the spec migration.

## Independent issues this PR had

Reviewed by `security-reviewer-deep` and `ad-tech-protocol-expert-deep`. Findings that would have blocked merge regardless of the supersession:

- **Wire leak of `agent_url`** (`permission_denied.py:240`): the PR ships `details.agent_url` in plaintext on recognized branches, directly contradicting its own indistinguishability claim. Tests at `test_tier2_parity_contract.py:191-205` actively pin this divergence.
- **`ADCP_PERMISSION_DENIED_BUDGET_MS=0` silently disables** the timing-oracle defense — no warning, no fallback.
- **`nan` returns nan** (no protection); **`inf` hangs dispatch indefinitely** (DoS via `asyncio.sleep(inf)`).
- **DEBUG-log path leaks** `scope` / `status` in plaintext when `audit_sink=None`, defeating the audit-hashing claim.
- **5 s sink timeout vs 50 ms budget** — 100× asymmetry reopens the timing oracle at a 5 s margin if the sink times out on one branch.
- **48-bit hash truncation is invertible** by operators with the candidate URL list — the "log-scraping cannot reconstruct" claim was overstated.
- **No transport-level integration test** through `ADCPAgentExecutor` or `build_mcp_error_result`. `_to_mcp` in `translate.py` leaks the discriminator via the `Details:` text line.

## Residual value moved to #772

The narrow follow-up at #772 captures what remains useful:

- Latency budget on the *two* remaining `PERMISSION_DENIED` branches (unrecognized + unknown-status default-reject), with the env-var hardening this PR was missing.
- Uniform `DenialAuditRecord` schema across all four denial branches (incl. the dedicated-code paths) — SecOps telemetry parity, not wire parity.
- HMAC-or-plaintext-with-documented-threat-model for the audit `agent_url` (rejects 48-bit truncation).
- Transport-level integration tests.
- Estimated ~150 LoC vs. this PR's 1,106.

The parent issue #392 will also close, since its 4-numbered design list is largely obsoleted by #748 and the residue is covered by #772.

Thanks to the original implementing agent for the careful scoping note — the "holding for human security review" comment was the right call, and the review surfaced the supersession.

bokelley:
Acknowledged — thanks for the detailed write-up. The supersession by #748 (dedicated wire codes + `recovery="terminal"`) and the residual capture in #772 make sense. The "holding for human security review" flag in this PR did its job; the reviewer findings you listed above are exactly the kind of blockers that gate should catch. No further action needed from triage on this branch.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01SV2WtK4kVkhhErSe8Eeind

---

PR #773, opened 2026-05-21
feat(decisioning): wire create_media_buy_store into PlatformHandler dispatch (closes #462)

### LINKED ISSUE #462: feat(server): createMediaBuyStore — opt-in targeting_overlay echo for property-lists / collection-lists sellers
Parent: #452

## JS source

6.7 commit `dda2a77e feat(server): createMediaBuyStore`.

## Use case

Narrow but mechanical. Sellers claiming `property-lists` / `collection-lists` specialisms have to echo `targeting_overlay` on `get_media_buys`. Currently hand-rolled in every adopter.

## Proposed API

```python
from adcp.decisioning import create_media_buy_store

platform.media_buy_store = create_media_buy_store(store)
```

Opt-in framework wiring that handles the echo automatically when the seller declares `property-lists` or `collection-lists`.

## Acceptance criteria

- `create_media_buy_store` exported from `adcp.decisioning`.
- `targeting_overlay` echoed correctly on `get_media_buys` responses for declared specialisms.
- No-op for sellers not claiming those specialisms.
- Tests cover both paths.

bokelley:
## Summary

Wires the already-merged `create_media_buy_store` helper into the framework dispatch path. Adopters now opt in via a kwarg on `create_adcp_server_from_platform` instead of hand-rolling `targeting_overlay` persistence + echo per adapter.

- **`create_media_buy`** — `persist_from_create` rides the existing `on_complete` seam, chained with the proposal v1.5 hook when both are wired, so both sync and HITL completions persist.
- **`update_media_buy`** — `merge_from_update` on `on_complete`, honoring the spec contract that `targeting_overlay` echoes from the most recent `create_media_buy` *or* `update_media_buy`.
- **`get_media_buys`** — `backfill` called inline before return.

New `_to_store_dict` helper normalizes Pydantic → JSON-compatible dict via `model_dump(mode='json', exclude_none=False)`. Adopters see the same shape that would go on the wire (`AnyUrl` → str, datetime → ISO-8601 str), and explicit nulls survive the merge contract (`None` = "clear this field").

## API shape decision

The triage comment surfaced an open question: kwarg on `create_adcp_server_from_platform` vs. post-construction `platform.media_buy_store = ...` (as the issue body proposed). Went with the kwarg — matches `buyer_agent_registry`, `webhook_sender`, `resource_resolver`, and every other framework opt-in.

```python
handler, executor, registry = create_adcp_server_from_platform(
    platform,
    media_buy_store=create_media_buy_store(adopter_store, capabilities=platform.capabilities),
)
```

## Test plan

- [x] 7 new integration tests in `tests/test_media_buy_store_integration.py` covering persist / merge / backfill through `PlatformHandler`, the no-op paths (store missing, seller lacks specialism), and a full create→get round-trip
- [x] 10 existing `tests/test_media_buy_store.py` unit tests still pass
- [x] Full test suite: 4881 passed, 0 regressions
- [x] `ruff check` + `mypy` clean on changed files
- [ ] Manual smoke through an adopter platform (defer to follow-up if needed)

Closes #462.

### DIFF
--- a/src/adcp/decisioning/handler.py
+++ b/src/adcp/decisioning/handler.py
@@ -186,6 +186,7 @@ from adcp.types import (
 if TYPE_CHECKING:
     from concurrent.futures import ThreadPoolExecutor
 
+    from adcp.decisioning.media_buy_store import MediaBuyStore
     from adcp.decisioning.platform import DecisioningPlatform
     from adcp.decisioning.property_list import PropertyListFetcher
     from adcp.decisioning.registry import BuyerAgent, BuyerAgentRegistry
@@ -806,6 +807,28 @@ def _extract_media_buy_id(result: Any) -> str | None:
     return str(value)
 
 
+def _to_store_dict(value: Any) -> Any:
+    """Normalize a Pydantic model OR plain dict to a JSON-compatible
+    dict for the :class:`MediaBuyStore` adopter contract.
+
+    The store Protocol uses ``Any`` typing, but the reference impl (and
+    the JS-side port) expects dict-shaped payloads — ``.get("packages")``,
+    ``["media_buy_id"]``, etc. Adopters writing against the Protocol
+    benefit from a stable shape regardless of whether their platform
+    method returned a typed Pydantic response or a hand-built dict.
+
+    ``mode='json'`` serializes ``AnyUrl`` to ``str`` and ``datetime``
+    to ISO-8601 ``str`` so the dict matches what would go over the wire
+    — adopters serializing to JSON for persistence don't have to
+    re-normalize. ``exclude_none=False`` preserves explicit nulls (the
+    merge contract treats ``None`` as "clear this field"; dropping them
+    would silently change semantics).
+    """
+    if hasattr(value, "model_dump"):
+        return value.model_dump(mode="json", exclude_none=False, by_alias=False)
+    return value
+
+
 class PlatformHandler(ADCPHandler[ToolContext]):
     """ADCPHandler subclass that routes wire requests to a
     :class:`DecisioningPlatform` via :func:`_invoke_platform_method`.
@@ -998,6 +1021,7 @@ class PlatformHandler(ADCPHandler[ToolContext]):
         buyer_agent_registry: BuyerAgentRegistry | None = None,
         config_store: ProductConfigStore | None = None,
         property_list_fetcher: PropertyListFetcher | None = None,
+        media_buy_store: MediaBuyStore | None = None,
         advertise_all: bool = False,
     ) -> None:
         super().__init__()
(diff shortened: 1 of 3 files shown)


bokelley:
Thanks @aao-ipr-bot — pushed `97e650ae` addressing the blocker and both flagged follow-ups.

## Blocker: credential leak

Replaced the blanket `_to_store_dict(params)` with three whitelist projection helpers in `handler.py`:

- `_project_create_request_for_store` — projects to `{packages: [{buyer_ref, targeting_overlay}, ...]}` and nothing else
- `_project_create_response_for_store` — projects to `{media_buy_id, packages: [{buyer_ref, package_id, targeting_overlay}, ...]}`
- `_project_update_patch_for_store` — projects to `{packages: [{package_id, targeting_overlay}, ...], new_packages: [{buyer_ref, package_id, targeting_overlay}, ...]}`

Whitelist > blacklist exactly as you flagged — future schema additions of credential-shaped fields can't reintroduce the leak. Pydantic's `model_dump(include=...)` does the projection, with parallel manual paths for dict inputs.

## Follow-ups addressed in the same commit

1. **Persist-failure / proposal-finalize ordering** — both `persist_from_create` and `merge_from_update` now run inside `try/except Exception → logger.exception`. Comment at the create-shim hook explains the ordering invariant: the create has already succeeded in the adapter and the proposal hook (if wired) has already marked the proposal CONSUMED, so unwinding via the exception path would surface a buyer-visible failure on a successful inventory commit.
2. **Post-backfill credential strip** — `get_media_buys` now re-runs `strip_credentials_from_wire_result("get_media_buys", result)` after backfill mutates the dict. Defense in depth against a buggy or malicious store re-injecting `governance_agents[].authentication` / `billing_entity.bank`.

## Regression tests

Added four tests in `tests/test_media_buy_store_integration.py`:

- `test_create_media_buy_does_not_leak_webhook_credentials_to_store` — asserts the persist call's `request_dict.keys()` is exactly `{"packages"}` and that the bearer token literal is nowhere in the flattened structure
- `test_update_media_buy_does_not_leak_webhook_credentials_to_store` — same shape check for the merge call's patch
- `test_create_media_buy_persist_failure_does_not_break_create_path` — uses a `_RaisingStore` to verify the create succeeds normally despite a store outage
- `test_get_media_buys_strips_credentials_re_injected_by_store` — uses a `_ReinjectingStore` that re-injects `governance_agents[].authentication` during backfill and asserts the field is gone in the returned response

## Other notes

- **Merge semantics in the fixture** (follow-up #4): the recording store's key-level `update()` mirrors the existing on-main reference impl in `tests/test_media_buy_store.py:214-237` (which also encodes key-merge via `_test_merge_from_update_preserves_prior_fields_when_patch_omits_them`). If the spec really mandates full replacement per `package-update.json`, the on-main reference impl is also wrong — happy to file a separate issue once we confirm the spec read, but I'd rather not change two test fixtures' semantics inline with a security fix.
- **Typed-vs-dict return contract** (follow-up #3): noted, leaving for a separate cleanup PR — same reasoning, this PR is already scope-creeping.
- **Nit on test response subscript**: added inline comments in the new credential-leak tests explaining the dict shape.

Suite: 4894 passed, 0 regressions. `ruff` + `mypy` clean.

Issue #711, opened 2026-05-13
feat: permissive resolver for bare authorized_agents entries with top-level properties


bokelley:
## Problem

The strict resolver (`get_properties_by_agent`) returns `[]` for a "bare" entry — `{url, authorized_for}` with no `authorization_type` and no selector. That's correct per the JSON Schema (every `oneOf` variant requires `authorization_type` + a selector). But it breaks operational reality.

Today, multiple real-world publishers ship this exact shape with a top-level `properties[]` block:

```json
{
  "$schema": "https://adcontextprotocol.org/schemas/v1/adagents.json",
  "authorized_agents": [
    {"url": "https://wonderstruck.sales-agent.scope3.com", "authorized_for": "Display banners"}
  ],
  "properties": [
    {"property_id": "main_site", "property_type": "website", "name": "Main site",
     "identifiers": [{"type": "domain", "value": "wonderstruck.org"}], "tags": ["sites"]}
  ]
}
```

Publisher intent: "this agent is authorized for everything I list." Strict resolver: `[]`. SDK consumers either:

1. Eat the regression — products can't bind to these publishers.
2. Reinvent permissive resolution locally — which everyone will do differently, defeating the SDK's job.

We just shipped #2 in salesagent (see [salesagent#377](https://github.com/bokelley/salesagent/issues/377)) but would rather depend on the SDK.

## Proposed API

A sibling helper that explicitly opts into the unbound interpretation, leaving `get_properties_by_agent` strict:

```python
from adcp.adagents import resolve_properties_for_agent

# Default: same as get_properties_by_agent (strict, spec-conformant)
props = resolve_properties_for_agent(adagents, agent_url)

# Permissive: bare entry + top-level properties → return top-level
props = resolve_properties_for_agent(adagents, agent_url, mode="permissive")
```

Or, if you prefer a separate function name to keep the call site obvious:

```python
from adcp.adagents import resolve_properties_for_agent_permissive
```

Semantics for `mode="permissive"`:

1. If the agent's entry has `authorization_type` → identical to strict path.
2. If the agent's entry is bare AND the file has a top-level `properties[]` array → return that array (caller can still domain-filter / property_type-filter as needed).
3. If the agent isn't in `authorized_agents[]` at all → still `[]` (no change).
4. If the agent's entry is bare AND there is no top-level `properties[]` → still `[]`.

[...]

PR #827, opened 2026-05-23
feat(adagents): add permissive property resolver

sangilish:
## Summary

Refs #711.

This adds an explicit opt-in resolver for operational adagents.json files that list a sales agent in `authorized_agents[]` but omit the schema-conformant `authorization_type` selector.

The existing `get_properties_by_agent()` behavior stays strict and unchanged. New callers can use:

```python
resolve_properties_for_agent(adagents_data, agent_url, mode="permissive")
```

In permissive mode, a matching bare `authorized_agents` entry resolves to the file's top-level `properties[]`. The fallback only applies when the entry has no `authorization_type` and no known selector field. If the agent is not listed, or if the entry has an explicit selector, the result remains the strict result.

## What changed

- Added `resolve_properties_for_agent(..., mode="strict" | "permissive")`.
- Kept `get_properties_by_agent()` as the strict path.
- Exported the new helper from `adcp.__all__` and updated the public API snapshot.
- Added tests covering:
  - default strict behavior for bare entries remains empty
  - permissive bare-entry fallback returns top-level properties
  - permissive mode still requires the agent URL to be listed
  - permissive mode does not override broken explicit selectors
  - revoked publisher domains are excluded from fallback results
  - unknown modes fail loudly

## Testing

- `uv run pytest tests/test_adagents.py -q -k "resolve_properties_for_agent or get_properties_by_agent"`
- `uv run pytest tests/test_public_api.py tests/test_adagents.py -q`
- `uv run pytest -q`
- `uv run ruff check src/adcp/adagents.py src/adcp/__init__.py tests/test_adagents.py tests/fixtures/public_api_snapshot.json`
- `uv run mypy src/adcp`
- `git diff --check`

Note: `uv run ruff check` across the full repository currently reports pre-existing lint drift in unrelated tests. The files touched in this PR pass ruff.

CI note: after marking this ready, the only failing GitHub check is code_review (Argus). It fails before review execution while minting the GitHub App token because the app client id / app id secret is unavailable in this fork PR context. This is not a code/test failure in this branch.


bokelley:
Superseded by #863, which ports this feature onto a repo-owned branch and adds the fail-closed hardening requested by expert review.

PR #939, opened 2026-06-07
feat(testing): reference UpstreamRecorder middleware for query_upstream_traffic

### LINKED ISSUE #347: feat(testing): reference upstream-traffic recorder middleware for query_upstream_traffic adopters (adcp#3830 item 4)
Python sibling of adcp-client#TBD. Ship a reference middleware Python adopters drop into their \`comply_test_controller\` implementation to populate the session-scoped recording buffer that \`query_upstream_traffic\` returns.

## Why

PR adcontextprotocol/adcp#3816 ships the spec contract for \`upstream_traffic\` storyboard checks. Python-based adopters (sales-house adapters, signal marketplaces, governance tools written in FastAPI / Flask) opt in by exposing \`query_upstream_traffic\` from \`comply_test_controller\`. The recorder primitive lets them avoid hand-rolling the buffer + redaction + caller-scoping that the spec requires.

## What to build

\`adcp.testing.upstream_recorder\`:

\`\`\`python
from adcp.testing import UpstreamRecorder

recorder = UpstreamRecorder(
    enabled=os.getenv(\"ADCP_SANDBOX\") == \"1\",
    redact_pattern=r\"^(authorization|token|api_key|...)\$\",
    buffer_size=1000,
    ttl_seconds=3600,
)

# Wire as httpx event hook / requests session adapter
import httpx
client = httpx.Client(event_hooks={\"request\": [recorder.on_request], \"response\": [recorder.on_response]})

# Or for requests:
import requests
requests.Session().mount(\"https://\", recorder.requests_adapter)

# In your comply_test_controller handler for scenario: query_upstream_traffic:
def handle_query_upstream_traffic(req, principal):
    result = recorder.query(
        principal=principal,
        since_timestamp=req.params.get(\"since_timestamp\"),
        endpoint_pattern=req.params.get(\"endpoint_pattern\"),
        limit=req.params.get(\"limit\", 100),
    )
    return {
        \"success\": True,
        \"recorded_calls\": result.items,
        \"total_count\": result.total,
        \"truncated\": result.truncated,
        \"since_timestamp\": req.params[\"since_timestamp\"],
    }
\`\`\`

Same key behaviors as the Node sibling: caller scoping (per-principal buffer), redaction at recording time, sandbox-only gating, optional purpose tagging callback.

## HTTP client coverage

Ship adapters for the common Python HTTP clients:
- \`httpx\` (event hooks — primary)
- \`requests\` (transport adapter)
- \`aiohttp\` (trace config)
- \`urllib3\` (lowest level — nice-to-have)

## References

- Spec contract: adcontextprotocol/adcp#3816 (merged)
- Sibling Node issue: adcp-client (file in parallel)
- Source issue: adcontextprotocol/adcp#3830 item 4
- Runner-side companion: adcp-client#1253 (consumer side; orthogonal to this)

bokelley:
Closes #347

## What

Adds a test-only `UpstreamRecorder` (`src/adcp/testing/upstream_recorder.py`) so adopters can answer the AdCP `query_upstream_traffic` conformance scenario — the mechanism that raises the bar against façade adapters that satisfy AdCP schema requirements with synthetic placeholders without forwarding data upstream.

## Design

Mirrors the JS reference recorder (adcp-client #1304) adapted to Python/httpx:

- **Primary adapter — httpx async event hooks.** `recorder.httpx_hooks()` returns an `event_hooks` dict for `httpx.AsyncClient`; the response hook captures each call once with its `status_code`.
- **Secondary clients** (`requests` / `aiohttp`) use the `record(...)` manual escape hatch — they have no native httpx-style hook surface.
- **Caller scoping** (spec-mandated). Records are keyed on the principal bound via `principal_scope` / `principal_scope_async` / `run_with_principal` (ContextVar, survives `await`). `query(principal=...)` returns only that principal's calls; cross-caller traffic is never returned.
- **Secret redaction at record time** using the normative `SECRET_KEY_PATTERN` from the `UpstreamTrafficSuccess` branch — recursive into JSON bodies and header maps, `[redacted]` literal.
- **Fail-closed attribution.** Calls recorded outside any principal scope are dropped (unattributable); empty/non-string principals raise `UpstreamRecorderScopeError` rather than silently matching nothing.
- `query(...)` returns an `UpstreamTrafficResult` whose `to_response_dict()` projects to the raw-attestation `UpstreamTrafficSuccess` shape (`success`, `recorded_calls`, `total_count`, `since_timestamp`, `truncated`).

Emits `attestation_mode="raw"` only — digest mode is left to adopters whose privacy policy requires it.

## Tests

`tests/test_upstream_recorder.py` (23 cases) records real httpx traffic via `httpx.MockTransport` (no live network) and validates each `recorded_calls[]` item against the inline item subschema from the rc.9 `comply-test-controller-response.json` cache. Covers capture, status code, per-principal scoping, drop-outside-scope, redaction, `endpoint_pattern` glob (anchored), `since_timestamp` / `limit` filtering, ordering, and payload truncation.

## Verification

`make lint`, `make typecheck`, and the new tests all pass. No `ADCP_VERSION` bump, no schema re-download. Write scope limited to the recorder, `src/adcp/testing/__init__.py`, and the test file.

### DIFF
--- a/src/adcp/testing/__init__.py
+++ b/src/adcp/testing/__init__.py
@@ -42,6 +42,12 @@ from adcp.testing.test_helpers import (
     test_agent_client,
     test_agent_no_auth,
 )
+from adcp.testing.upstream_recorder import (
+    RecordedCall,
+    UpstreamRecorder,
+    UpstreamRecorderScopeError,
+    UpstreamTrafficResult,
+)
 
 __all__ = [
     "AdcpErrorPayload",
@@ -64,4 +70,8 @@ __all__ = [
     "TEST_AGENT_MCP_NO_AUTH_CONFIG",
     "TEST_AGENT_A2A_NO_AUTH_CONFIG",
     "CREATIVE_AGENT_CONFIG",
+    "UpstreamRecorder",
+    "UpstreamRecorderScopeError",
+    "UpstreamTrafficResult",
+    "RecordedCall",
 ]
--- a/src/adcp/testing/upstream_recorder.py
+++ b/src/adcp/testing/upstream_recorder.py
@@ -1,0 +1,496 @@
+"""Reference upstream-traffic recorder for ``query_upstream_traffic`` conformance.
+
+The AdCP ``query_upstream_traffic`` scenario asks a seller to report the
+outbound HTTP calls it made while servicing a buyer's request — the
+mechanism that raises the bar against façade adapters that satisfy AdCP
+schema requirements with synthetic placeholders without forwarding data
+upstream (see the ``UpstreamTrafficSuccess`` branch of
+``comply-test-controller-response.json``).
+
+:class:`UpstreamRecorder` is a test-only helper adopters can wire into
+their httpx-based upstream client so a ``comply_test_controller``
+handler can answer ``query_upstream_traffic`` with a spec-shaped
+``recorded_calls`` array.
+
+Primary adapter — httpx async event hooks::
+
+    recorder = UpstreamRecorder()
+    async with recorder.principal_scope("agent.example.com"):
+        client = httpx.AsyncClient(event_hooks=recorder.httpx_hooks())
+        await client.post("https://api.example.com/v2/audience/upload", json=...)
+
+    result = recorder.query(principal="agent.example.com")
+    response = result.to_response_dict()  # shape for query_upstream_traffic
+
+Secondary clients (``requests`` / ``aiohttp``) have no native httpx-style
+hook surface; adopters call :meth:`UpstreamRecorder.record` directly from
+their own client wrapper as a manual escape hatch.
+
+Spec obligations this recorder implements:
+
+* **Caller scoping.** Records are keyed on the principal bound via
+  :meth:`principal_scope` / :meth:`run_with_principal`. :meth:`query`
+  requires a principal and returns only that principal's calls —
+  cross-caller traffic is never returned regardless of ``since_timestamp``.
+* **Secret redaction.** Values at keys matching the normative
+  case-insensitive :data:`SECRET_KEY_PATTERN` are recursively replaced
+  with ``[redacted]`` at record time, in both header maps and decoded
+  JSON request bodies.
+* **Fail-closed scoping.** Recording outside any principal scope drops
+  the call (it is unattributable); an empty / non-string principal raises
+  :class:`UpstreamRecorderScopeError` rather than silently matching nothing.
+"""
+
+from __future__ import annotations
+
+import json
+import re
+from contextlib import asynccontextmanager, contextmanager
+from contextvars import ContextVar
+from dataclasses import dataclass
+from datetime import datetime, timezone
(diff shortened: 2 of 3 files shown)


bokelley:
Pushed the redaction fixes addressing the CHANGES_REQUESTED review (`b6fe4563` on `claude/issue-347-upstream-recorder`).

**Blocking fixes**

1. **URL query-string secrets (High).** `record()` now runs the url through `_redact_url` before storage: `urlsplit` → `parse_qsl` → key-redact → rebuild. The redacted url feeds both `url` and `endpoint` (and `host`/`path`), so query-param auth (`?api_key=`, `?access_token=`) and presigned-URL tokens no longer reach the wire. Added `URL_QUERY_SECRET_PATTERN` (`X-Amz-Signature` / `Signature` / `X-Goog-Signature`) matched only against query keys, so the normative body pattern is unchanged.

2. **Form-urlencoded bodies (High).** `_normalize_payload` now detects `application/x-www-form-urlencoded`, parses via `parse_qsl`, key-redacts with the normative `SECRET_KEY_PATTERN`, and re-encodes. `refresh_token`/`client_secret` are redacted; non-secret keys (`grant_type`, `audience`) survive.

3. **Truncated `payload_length` (correctness).** Both truncation paths now route through `_truncate_string`, which reserves room for the `[…truncated]` marker (so the emitted payload stays within the schema `maxLength` 65536, :505) and reports the UTF-8 byte length of the **emitted** value per :515. `test_oversized_string_payload_is_truncated` now asserts `payload_length == len(emitted)` rather than the original length.

**Nits**

- Removed the dead `_redact_headers(headers)` call and corrected the module docstring — it now states the recorder covers JSON bodies, form bodies, and URL query (not header maps; no header reaches the wire).
- Wrapped `request.content` in `try/except httpx.RequestNotRead` → `payload=None` for streaming bodies.
- Purpose reconstruction uses `dataclasses.replace(call, purpose=purpose)`.

**Tests** (tests/test_upstream_recorder.py): parametrized `?api_key=`/`?access_token=`/`?X-Amz-Signature=` redaction asserting the secret is gone from both `url` and `endpoint` while non-secret params survive; form-urlencoded `refresh_token`/`client_secret` redaction; and a 70 KB truncation case that schema-validates the emitted item with `payload_length == len(emitted)` and `len(payload) <= 65536`.

`uv run pytest tests/test_upstream_recorder.py -q`: **28 passed**. ruff + mypy + bandit pass via pre-commit.

bokelley:
Noted — the three blockers from the CHANGES_REQUESTED review (URL query-string redaction, form-urlencoded body redaction, truncated `payload_length`) and all four nits are addressed in `b6fe4563`. 28 tests passing with ruff/mypy/bandit clean. Ready for re-review by the reviewer who requested changes.

---

bokelley:
Acknowledged — status noted. The PR is in the reviewer's hands for re-review; no triage action needed here.

---

PR #959, opened 2026-06-27
perf(types): defer pydantic core-schema builds to cut import memory

andybevan-scope3:
## Why

`import adcp` builds the pydantic-core validator and serializer for every one of the ~1500 generated model classes at import time, even though a given process only ever touches a handful. Those eager builds dominate the import-time memory footprint, and on the embedded sales agent (2Gi limit) they push idle RSS high enough to OOM in staging.

## What changed (in plain terms)

**Build each model only when it's actually used, not when `adcp` is imported.**

- Added `defer_build=True` to `AdCPBaseModel`. Pydantic now skips building a model's validator/serializer at class-definition time and builds it lazily the first time that model is validated or serialized. After a bare `import adcp`, only **18 of 1564** model classes are built; the other 1546 stay deferred until something uses them.

- Fixed one interaction this exposed. `AdCPBaseModel.model_dump` already forces `serialize_as_any=True` (so subclass serializers fire when nested under a base-typed field). That option makes pydantic serialize each nested value using its *own* class serializer, and a model that only ever appears as a nested field never gets its lazy build triggered, so that serializer is still a placeholder. Serializing then raised `TypeError: 'MockValSer' object is not an instance of 'SchemaSerializer'`. The fix: when that specific error occurs, walk the object graph and build the still-deferred nested classes, then retry. This builds only the classes that appear in real payloads, so the memory win is preserved.

Net effect: same validation and serialization behavior as before, just with the model-build cost moved off the import path and paid lazily per model. The deferred-build memory reclaim is large on the Linux deploy target (the OOM case); it's smaller on macOS because of allocator differences.

## Tests

- Full suite passes: **5804 passed**, 41 skipped, 1 xfailed (pre-existing, unrelated).
- Added `test_serialize_as_any_builds_deferred_nested_serializer` pinning the nested-deferred-serializer path (a child model created only via the parent's `model_validate`, then dumped) — this is exactly the case that broke before the fix.
- `make lint` and `make typecheck-all` clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,10 +6,10 @@ on:
   pull_request:
     branches: [main]
 
-# Default @adcp/sdk runner alias for non-matrix storyboard jobs.
-# The reference seller storyboard job below runs both sticky 3.0 and 3.1 tags.
+# Default @adcp/sdk runner alias for storyboard jobs. Tracks the current
+# stable @adcp/sdk release via the ``latest`` npm dist-tag.
 env:
-  ADCP_SDK_VERSION: "adcp-3.1"
+  ADCP_SDK_VERSION: "latest"
 
 concurrency:
   group: ci-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


bokelley (approved):
Reviewed the deferred pydantic build change and ran focused local coverage: tests/test_serialize_as_any_default.py, tests/test_type_coercion.py, tests/test_response_builder_subclass.py, tests/test_rootmodel_proxy.py, and tests/test_extra_policy.py all passed locally (73 tests). CI is green.

PR #960, opened 2026-06-27
chore(types): prune unused bundled generated modules

andybevan-scope3:
## What this does, in plain terms

The library auto-generates Python type files from JSON schemas into
`src/adcp/types/generated_poc/`. One sub-folder, `bundled/`, had grown to
dominate that tree: ~92% of all generated classes lived there.

That folder is generated from "bundled" schemas, which inline their entire
reference graph. So every bundled message file re-emitted its own private copy
of the same shared sub-schemas (provenance, verification, assets, etc.) over and
over (e.g. `VerifyAgent` appeared 3,762 times).

The catch: **almost none of it is used.** Of ~97 bundled modules, exactly **one**
is imported by real code: `bundled/protocol/get_adcp_capabilities_response.py`
(via `adcp.types.capabilities`, whose typed sub-models only exist in the inlined
form). `consolidate_exports.py` already excludes the bundled tree from the public
namespace, and it isn't even loaded at `import adcp`.

So this PR deletes the 95 unreferenced bundled modules and teaches the generator
not to recreate them.

## What changed

- **Deleted** 95 unused `bundled/` modules (kept only the one consumed module +
  its package `__init__` files).
- **Added** a `prune_unused_bundled_modules()` step to `scripts/generate_types.py`
  (runs after post-generation fixes, before consolidation). Generation itself is
  unchanged; the step only removes the unreferenced output afterward, so every
  retained module stays byte-identical. Keep-set lives in the `BUNDLED_KEEP`
  constant.

## Impact

| | Before | After |
|---|---|---|
| Generated class definitions | 30,173 | 2,382 (−27,791, −92%) |
| Generated source lines | ~686k | ~61k (−625k) |
| Full test suite | 5,824 passed | 5,824 passed, 0 failed |

`make lint`, `make typecheck`, and the full `pytest` suite all pass.

## Scope notes

- This is a **source-tree / class-count** cleanup. It does **not** lower
  `import adcp` memory (the bundled tree was lazy and never loaded). Import-time
  RSS is a separate concern handled by the `defer_build` work in the el-paso
  worktree.
- Follow-up (not in this PR): regeneration still generates `bundled/media_buy`
  and then prunes it. That split-generation pass is now dead weight and can be
  skipped to speed up `make regenerate-schemas`.
- A full `make regenerate-schemas` run to exercise the generation→prune
  integration end-to-end has not been run yet; the end-state and the prune
  function were verified directly (deletion + full suite green, idempotent prune).


bokelley (approved):
Reviewed the generated bundled-module pruning and live import surface. Local focused checks passed: tests/test_code_generation.py, tests/test_decisioning_capabilities_submodule.py, and tests/test_asset_aliases_stable.py all passed (89 tests), plus a syntax/import smoke for _generated.py and the retained bundled capabilities module. I also attempted a full local scripts/generate_types.py run; it remained active in datamodel-code-generator for ~13 minutes, so I stopped it and relied on the successful CI schema-validation job for the full regeneration environment.

PR #961, opened 2026-06-27
ci: point storyboard runner at @adcp/sdk latest instead of adcp-3.1

andybevan-scope3:
## Why

Every \`AdCP storyboard runner\` job is failing on \`npm error notarget No matching version found for @adcp/sdk@adcp-3.1\`. This is not a defect in any PR — the \`adcp-3.1\` npm dist-tag was removed from the registry.

Root cause (traced in \`adcontextprotocol/adcp-client\`): commit \`509ed55b fix: publish stable releases to latest\` (2026-06-26) changed the SDK release process. It previously auto-published every stable release under a protocol-derived tag (\`adcp-3.1\`); it now publishes stable under \`latest\` and no longer emits \`adcp-3.1\`. AdCP 3.1 is the current stable line, shipped as \`@adcp/sdk@latest\` (currently 9.3.0). The SDK's \`NPM-DIST-TAGS.md\` now states an \`adcp-3.X\` compat tag exists only once that line becomes a *maintenance* line behind a newer stable — so there is intentionally no \`adcp-3.1\` tag today.

This passed on main as recently as June 20 (when the tag still existed) and started failing for all PRs after June 26. It blocks every PR because the storyboard jobs are required checks.

## What changed

- \`ADCP_SDK_VERSION\` default and the seller_agent.py storyboard matrix leg now track \`latest\` (which *is* the AdCP 3.1 GA line) instead of the removed \`adcp-3.1\` tag.
- Kept the sticky \`adcp-3.0\` tag for backwards-compatibility coverage.
- Updated the stale comments that described \`adcp-3.1\` as the current-line tag.

## Tradeoff

\`latest\` is a moving target, so a future \`@adcp/sdk\` release could change storyboard behavior without a tag bump here. That is acceptable (and arguably desirable — we want to know): the sticky \`adcp-3.0\` leg still provides a fixed backwards-compat gate. If the org later wants a stable \`adcp-3.1\` compat tag, that is a separate change in the SDK repo's publish process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -361,12 +361,14 @@ jobs:
     name: AdCP storyboard runner — examples/seller_agent.py (@adcp/sdk ${{ matrix.adcp-sdk-tag }})
     runs-on: ubuntu-latest
     # Blocking gate: examples/seller_agent.py is the Python-owned
-    # reference target for bidirectional storyboard interop. ``latest``
+    # reference target for bidirectional storyboard interop. The matrix
+    # runs two legs: the sticky ``adcp-3.0`` tag is a fixed, reproducible
+    # backwards-compat floor (the AdCP 3.0 runner line), while ``latest``
     # tracks the current stable @adcp/sdk release (the AdCP 3.1 line).
     strategy:
       fail-fast: false
       matrix:
-        adcp-sdk-tag: ["latest"]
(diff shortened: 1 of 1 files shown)


bokelley (approved):
Reviewed the workflow diff and verified current npm dist-tags locally: @adcp/sdk publishes adcp-3.0 and latest, with no adcp-3.1 tag. CI is green.

PR #974, opened 2026-07-27
fix(signing): block CGNAT and 6to4 relay ranges in SSRF validation

### LINKED ISSUE #973: SSRF validation accepts RFC 6598 CGNAT (100.64.0.0/10), which AdCP 3.1.1 requires blocking
## Summary

`adcp.signing.resolve_and_validate_host` accepts destinations in `100.64.0.0/10`. AdCP 3.1.1 lists that range explicitly in the reserved-range deny list a fetcher **MUST** apply before any outbound fetch to a counterparty-controlled URL.

Spec: `dist/docs/3.1.0/building/by-layer/L1/security.mdx`, "Webhook URL validation (SSRF)", step 2 — *"IPv4: RFC 1918 (...), RFC 6598 CGNAT (`100.64.0.0/10`), loopback (...)"*.

## Cause

The validator classifies reserved addresses with Python `ipaddress` flags:

```python
ip.is_private or ip.is_loopback or ip.is_link_local
or ip.is_multicast or ip.is_reserved or ip.is_unspecified
```

`is_private` is `False` across `100.64.0.0/10` — RFC 6598 designates *shared* address space, not private space, so Python does not classify it. No other flag covers it either.

## Reproduce

```python
from unittest.mock import patch
from adcp.signing import resolve_and_validate_host

with patch("adcp.signing.jwks.socket.getaddrinfo",
           return_value=[(2, 1, 6, "", ("100.64.0.1", 0))]):
    print(resolve_and_validate_host("https://buyer-supplied.example/hook"))
# ("buyer-supplied.example", "100.64.0.1", 443)  -- accepted
```

## Impact

CGNAT space is routable inside carrier networks and is commonly used by container/overlay networks, so a buyer-supplied `push_notification_config.url` or `jwks_uri` resolving there reaches an operator internal address. It is pre-auth-reachable on the JWKS path.

Partially masked today: Alibaba metadata `100.100.100.200` sits inside this range and is caught by name via `BLOCKED_METADATA_IPS`. Every other address in the /10 is reachable.

## Also affected

`192.88.99.0/24` (RFC 7526 deprecated 6to4 relay anycast) falls through the same flag set. Traffic there is tunnelled by whichever relay answers, so the destination is unattributable.

## Fix

PR follows: adds an explicit `_EXTRA_BLOCKED_NETWORKS` tuple for the ranges the flags miss, behind the existing `allow_private` gate.

KonstantinMirin:
Fixes #973.

## What

`resolve_and_validate_host` classifies reserved addresses with Python's `ipaddress` flags plus an explicit cloud-metadata set. Two ranges fall through every flag:

| Range | Why the flags miss it | Why it matters |
|---|---|---|
| `100.64.0.0/10` | `is_private` is **False** — RFC 6598 designates *shared* address space, not private space | AdCP 3.1.1 names it in the deny list a fetcher **MUST** apply (`building/by-layer/L1/security.mdx`, "Webhook URL validation (SSRF)", step 2). Routable in carrier and container networks |
| `192.88.99.0/24` | `is_reserved` does not cover it | RFC 7526 deprecated 6to4 relay anycast — traffic is tunnelled by whichever relay answers, so the destination is unattributable |

So the SDK was accepting counterparty destinations that its own pinned spec requires it to reject.

## How

Adds `_EXTRA_BLOCKED_NETWORKS` and folds it into the existing condition. Deliberately **not** a rewrite to a full explicit prefix table — the flag-based approach tracks Python's IANA updates for free and is the right default; this only patches the two documented holes in it.

The new ranges sit behind the same `allow_private` gate as every other reserved range, so on-prem and test deployments keep their escape hatch. The cloud-metadata list stays unconditional.

## Interaction worth noting

Alibaba's metadata IP `100.100.100.200` lives *inside* `100.64.0.0/10`. It was previously reachable-by-name only through `BLOCKED_METADATA_IPS`; it is now covered twice. A regression test pins that the metadata check still wins under `allow_private=True`, so the escape hatch cannot unblock metadata.

## Tests

`test_ssrf_blocks_ranges_python_flags_miss` — parametrised over both CGNAT bounds, a mid-range address, and the 6to4 relay range. Verified red before the fix (4 failed) and green after.

Plus two behaviour-preservation guards:
- `test_ssrf_cgnat_honours_allow_private_override`
- `test_ssrf_alibaba_metadata_blocked_despite_allow_private`

Full suite: **5874 passed, 41 skipped, 1 xfailed** — no regressions. `ruff format`, `ruff check`, and `mypy` clean on both files.

## Compatibility

No public API change. Purely additive rejection of address space that is never a legitimate webhook or JWKS destination — hence `fix:`, not `fix!:`.

### DIFF
--- a/src/adcp/signing/jwks.py
+++ b/src/adcp/signing/jwks.py
@@ -55,6 +55,24 @@ BLOCKED_METADATA_IPS: frozenset[str] = frozenset(
     }
 )
 
+# Reserved ranges that Python's ``ipaddress`` flags do NOT classify, so the
+# ``is_private``/``is_reserved``/... check below misses them on its own.
+#
+# * ``100.64.0.0/10`` — RFC 6598 carrier-grade NAT. AdCP names this range
+#   explicitly in the reserved-range deny list a fetcher MUST apply
+#   (spec 3.1.1, "Webhook URL validation (SSRF)", step 2), but
+#   ``is_private`` is False for it: RFC 6598 designates shared address
+#   space, not private space. It is routable inside carrier and container
+#   networks (and is where Alibaba's metadata endpoint lives), so leaving
+#   it open is a real SSRF path into an operator's internal network.
+# * ``192.88.99.0/24`` — RFC 7526 deprecated 6to4 relay anycast. Traffic
+#   sent here is tunnelled by whatever relay answers, which makes the
+#   destination unattributable; ``is_reserved`` does not cover it.
+_EXTRA_BLOCKED_NETWORKS: tuple[ipaddress.IPv4Network | ipaddress.IPv6Network, ...] = (
+    ipaddress.ip_network("100.64.0.0/10"),
+    ipaddress.ip_network("192.88.99.0/24"),
+)
+
 # Recommended destination ports for hardened SSRF-validated outbound HTTP
 # deployments. AdCP itself does not constrain ``pushNotificationConfig.url``
 # ports (see ``schemas/cache/core/push-notification-config.json``), so the
@@ -272,6 +290,7 @@ def resolve_and_validate_host(
             or ip.is_multicast
             or ip.is_reserved
             or ip.is_unspecified
+            or any(ip in network for network in _EXTRA_BLOCKED_NETWORKS)
         ):
             last_rejection = f"resolved IP {ip} is in a reserved range"
             # Historical behavior of validate_jwks_uri was to raise on
--- a/src/adcp/signing/jwks.py
+++ b/src/adcp/signing/jwks.py
@@ -68,9 +68,35 @@ BLOCKED_METADATA_IPS: frozenset[str] = frozenset(
 # * ``192.88.99.0/24`` — RFC 7526 deprecated 6to4 relay anycast. Traffic
 #   sent here is tunnelled by whatever relay answers, which makes the
 #   destination unattributable; ``is_reserved`` does not cover it.
+# * ``192.31.196.0/24`` (RFC 7535 AS112-v4), ``192.52.193.0/24`` (RFC 7450
+#   AMT), ``192.175.48.0/24`` (RFC 7534 AS112 direct delegation) and
+#   ``2001:20::/28`` (RFC 7343 ORCHIDv2) — IANA special-use anycast and
+#   non-routable identifier space. None is ever a legitimate JWKS or webhook
+#   destination, and none is caught by any flag.
+#
+# Verified empirically across the supported interpreter matrix (3.10, 3.11,
+# 3.12, 3.13): every range above is classified non-reserved by all six flags on
+# ALL of them, so each entry is load-bearing on every supported version — none
+# is redundant anywhere. In particular ``100.64.0.0/10`` is still
+# ``is_private == False`` on 3.12.9 and 3.13.11.
+#
+# ``not ip.is_global`` is NOT a usable substitute: it reports True (i.e.
+# globally reachable) for the 6to4-relay, AS112, AMT and ORCHIDv2 ranges on
+# every supported version, so it would close none of these holes.
+#
+# Ranges deliberately absent because the flags already cover them on every
+# supported version — 6to4 ``2002::/16``, Teredo ``2001::/32``, NAT64
+# ``64:ff9b::/96``, and IPv4 NAT64 ``192.0.0.8``. Their embedded IPv4 needs no
+# decoding: the whole prefix is rejected before any address is unwrapped.
+# ``test_ssrf_flag_covered_special_ranges_stay_blocked`` pins that so a future
+# CPython reclassification cannot silently open them.
 _EXTRA_BLOCKED_NETWORKS: tuple[ipaddress.IPv4Network | ipaddress.IPv6Network, ...] = (
     ipaddress.ip_network("100.64.0.0/10"),
     ipaddress.ip_network("192.88.99.0/24"),
+    ipaddress.ip_network("192.31.196.0/24"),
+    ipaddress.ip_network("192.52.193.0/24"),
+    ipaddress.ip_network("192.175.48.0/24"),
+    ipaddress.ip_network("2001:20::/28"),
 )
 
 # Recommended destination ports for hardened SSRF-validated outbound HTTP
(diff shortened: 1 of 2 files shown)


KonstantinMirin:
Thanks for the careful read — the fail-closed ordering and metadata-precedence checks are exactly the parts I most wanted a second pair of eyes on.

Both nits are addressed in b6b3f47f:

- **v6-against-v4-network is now tested.** Two public IPv6 cases (`2606:4700:4700::1111`, `2001:4860:4860::8888`) that must pass rather than raise. You're right that the no-crash claim rested on CPython behaviour rather than a test — and since that behaviour is load-bearing for every v6 resolution, it deserved pinning.
- **`192.88.99.0/24` now has bounds** (`.0` and `.255`) plus an `allow_private` override case, so the 6to4 range is pinned directly rather than inferred from CGNAT's coverage.

On the NAT64/6to4 follow-up — I don't think there's an issue to file, and I'd rather flag that than leave a phantom vulnerability on the backlog. Both prefixes are already blocked wholesale, so there's nothing for the validator to decode:

```python
>>> ipaddress.ip_address('64:ff9b::a9fe:a9fe').is_private   # NAT64 -> 169.254.169.254
True
>>> ipaddress.ip_address('2002:a9fe:a9fe::').is_private     # 6to4  -> 169.254.169.254
True
```

Through the real validator, with `getaddrinfo` patched to return each address:

```
blocked  NAT64 -> 169.254.169.254   :: resolved IP 64:ff9b::a9fe:a9fe is in a reserved range
blocked  NAT64 -> 10.0.0.1          :: resolved IP 64:ff9b::a00:1 is in a reserved range
blocked  6to4  -> 169.254.169.254   :: resolved IP 2002:a9fe:a9fe:: is in a reserved range
blocked  6to4  -> 10.0.0.1          :: resolved IP 2002:a00:1:: is in a reserved range
```

Python classifies all of `64:ff9b::/96` (RFC 6052) and `2002::/16` (RFC 3056) as private, so the address is rejected before any socket opens regardless of what IPv4 it embeds. The `ipv4_mapped` unwrap at `jwks.py:264-265` is needed because `::ffff:0:0/96` is *not* covered by the flags — that asymmetry is what makes the tunnel prefixes look like they need the same treatment when they don't.

Worth noting this is the same class of gap the PR fixes, just in the other direction: CGNAT needed an explicit entry precisely because `is_private` is False for it, whereas these two are already True. The distinction is whether Python classifies the prefix, not whether an IPv4 is embedded.

Happy to add the tunnel addresses as regression cases if you'd like them pinned against a future CPython change to those classifications — that'd be a cheap guard, and unlike the decode approach it tests the property we actually depend on.

On CI: the one red job is `AdCP storyboard runner — v3 reference seller (translator)`, failing at `Seller failed to start within 30s` during boot. This diff touches only `src/adcp/signing/jwks.py` and its test file; the seller's boot path talks to a mock on `127.0.0.1`, which is rejected by `is_loopback` and unaffected by either added range. I've re-run it.

KonstantinMirin:
Follow-up and nit addressed in 61e6d926. I measured the classification across the full supported matrix (3.10.20, 3.11.13, 3.12.9, 3.13.11) rather than reasoning about it, which split the follow-up into four real gaps and three false alarms.

**Added** — non-reserved under all six flags on every supported version:

| Range | RFC |
|---|---|
| `192.31.196.0/24` | 7535 AS112-v4 anycast |
| `192.52.193.0/24` | 7450 AMT anycast |
| `192.175.48.0/24` | 7534 AS112 direct delegation |
| `2001:20::/28` | 7343 ORCHIDv2 |

The ORCHIDv2 entry also resolves nit 1 as a side effect — the `IPv4Network | IPv6Network` annotation is now accurate rather than forward-looking.

**Not added**, because the flags already reject them on 3.10/3.11/3.12/3.13: 6to4 `2002::/16`, Teredo `2001::/32`, and IPv4 NAT64 `192.0.0.8`. Same for `64:ff9b::/96` from the previous round. Their embedded IPv4 needs no decoding — the whole prefix is rejected before anything is unwrapped.

That block does depend on CPython's classification rather than on our own list, though, which is worth pinning. `test_ssrf_flag_covered_special_ranges_stay_blocked` now covers all six, so a future reclassification surfaces as a failing test instead of a silent hole. That felt like the right form for both this follow-up and the NAT64/6to4 one from the earlier round: test the property we actually depend on, rather than add a decode path for a bypass that doesn't exist.

Two corrections, both measured:

**`100.64.0.0/10` is not fixed on newer interpreters.** It is `is_private == False` on 3.12.9 and 3.13.11:

```
py3.12.9   CGNAT  flags_block=False  is_global=False
py3.13.11  CGNAT  flags_block=False  is_global=False
```

So the entry is load-bearing on every supported version. I did not add the suggested "redundant-but-harmless on 3.11.9+/3.12.4+" comment — it would have recorded something untrue in the source, and a future reader might have deleted the entry on the strength of it. (gh-113171 did land in that window, but it moved `is_global`, not `is_private`, and for other ranges.)

**`not ip.is_global` will not work as a substitute gate.** It reports `True` — globally reachable — for the 6to4-relay, AS112, AMT and ORCHIDv2 ranges on every supported version:

```
6to4relay     flags_block=False  is_global=True
AS112-v4      flags_block=False  is_global=True
AMT           flags_block=False  is_global=True
ORCHIDv2      flags_block=False  is_global=True
```

It would close none of the holes it was proposed for. The explicit list is not merely "defensible" here — it is the only option of the two that works.

One thing your review prompted that I would not have caught: the parametrised cases were building `AF_INET` 2-tuple sockaddrs for every address, so the new IPv6 entry would have exercised a record `getaddrinfo` never actually returns. There is now a helper picking the shape by family.

Red-before/green-after verified on the four new ranges (exactly those four fail with the entries removed). Full suite 5889 passed, 41 skipped, 1 xfailed. `ruff format`, `ruff check`, `mypy` clean.

bokelley:
Measurement-driven approach was the right call — the `100.64.0.0/10` finding (still `is_private == False` on 3.12.9/3.13.11) and the `is_global` gate failure across all four ranges would both have been invisible without running the full matrix. Testing the CPython classification property directly in `test_ssrf_flag_covered_special_ranges_stay_blocked` is the correct guard: a future reclassification surfaces as a failing test rather than a silent hole, and it pins what we actually depend on without adding decode paths for bypasses that don't exist.

---

Issue #979, opened 2026-07-28
Canonicalization drops a trailing empty query, so /p? and /p produce the same signature base


KonstantinMirin:
Found while vendoring the complete AdCP 3.1.1 request-signing vector set (#975). Reproduction is a `canonicalization.json` case, currently strict-xfail.

## Problem

A URL with a trailing `?` and an empty query loses the `?` during canonicalization.

| case | input | expected `@target-uri` | actual |
|---|---|---|---|
| `trailing-empty-query-preserved` | `https://seller.example.com/p?` | `https://seller.example.com/p?` | `https://seller.example.com/p` |

## Cause

`canonicalize_target_uri()` round-trips through `urlsplit` / `urlunsplit`:

```python
return urlunsplit((scheme, netloc, path, parts.query, ""))
```

`urlsplit` maps both `/p?` and `/p` to `query == ""`, and `urlunsplit` emits no `?` for an empty query. The distinction between "empty query" and "no query" is erased, so the two URLs canonicalize identically.

## Why it matters

RFC 9421 §2.2.2 derives `@target-uri` from the absolute request target, where `/p?` and `/p` are distinct. A counterparty that preserves the trailing `?` computes a different signature base than the SDK, and the signature fails to verify with no diagnostic beyond `request_signature_invalid` — the failure mode `canonicalization.json` exists to catch early.

Narrow, but it is a real divergence and the fix is contained: preserve the `?` when the original URL had one, e.g. by checking for `"?" in url` after the authority rather than relying on `urlunsplit` to round-trip it.

bokelley:
## Triage

**Classification:** Bug
**Bucket(s):** signing (note: `signing` label is not in this repo — bucket recorded in comment only)
**Status:** deferred

**Blocked-on: #980 — resurfaces on merge.**

PR #980 (by the same author, open as of today) vendors the complete AdCP 3.1.1 vector set and explicitly marks this case as a strict-xfail with a back-reference to this issue. The fix to `canonicalize_target_uri()` in `src/adcp/signing/canonical.py` should arrive in a follow-up PR *after* #980 merges, so the implementation PR can retire the xfail marker cleanly rather than landing against a test infrastructure that doesn't yet exist.

[...]

Issue #978, opened 2026-07-28
Canonicalization accepts five malformed authority shapes the spec requires be rejected; request_target_uri_malformed unimplemented


KonstantinMirin:
Found while vendoring the complete AdCP 3.1.1 request-signing vector set (#975). Reproductions are `canonicalization.json` cases, currently strict-xfail in `tests/conformance/signing/test_canonicalization.py`.

## Problem

The spec requires signers to refuse to sign, and verifiers to reject, URLs whose authority is syntactically malformed:

> If the authority has userinfo but no host (`https://user@/p`), no host at all (`https://:443/p` or `https:///p`), a bracketed host missing a closing bracket (`https://[::1/p`), or a bare IPv6 address outside brackets (`https://fe80::1/p`), the URL is malformed; signers MUST NOT sign it and verifiers MUST reject with `request_target_uri_malformed`.
> — `security.mdx`, `@target-uri` canonicalization step 3

> **IPv6 zone identifiers (RFC 6874) MUST be rejected.** ... verifiers MUST reject `@target-uri` values containing `%25` inside `[...]` with `request_target_uri_malformed`.
> — ibid., step 2

`canonicalize_target_uri()` rejects only one of the six shapes. The rest are canonicalized into a well-formed-looking result and signed.

## Failing cases

| case | input | expected | actual |
|---|---|---|---|
| `malformed-port-without-host` | `https://:443/p` | reject | accepted as `https:///p` |
| `malformed-userinfo-without-host` | `https://user@/p` | reject | accepted as `https:///p` |
| `malformed-empty-authority` | `https:///p` | reject | accepted as `https:///p` |
| `malformed-bare-ipv6` | `https://fe80::1/p` | reject | accepted as `https://fe80::1/p` |
| `malformed-ipv6-zone-identifier` | `https://[fe80::1%25eth0]/p` | reject | accepted unchanged |
| `malformed-ipv6-missing-closing-bracket` | `https://[::1/p` | reject | ✅ rejected (`ValueError`) |

The first three collapse to an **empty authority**, so `@authority` enters the signature base as the empty string and `@target-uri` as `https:///p`. Two different malformed inputs (`https://:443/p` and `https://user@/p`) canonicalize to the identical base — signatures become transferable between them.

[...]

Issue #977, opened 2026-07-28
@target-uri canonicalization applies no IDNA A-label conversion, and the verifier does not reject raw U-labels


KonstantinMirin:
Found while vendoring the complete AdCP 3.1.1 request-signing vector set (#975). Reproductions are newly vendored cases, currently strict-xfail.

## Problem

`@target-uri` / `@authority` canonicalization in `src/adcp/signing/canonical.py` performs no IDN processing. `_canon_authority()` lowercases the host with `str.lower()` and returns it as-is:

```python
host = host.lower()
if port is not None and port != _DEFAULT_PORTS.get(scheme):
    return f"{host}:{port}"
return host
```

The SDK **already ships** the correct helper — `adcp.signing._idna_canonicalize.canonicalize_host()` — and wires it into `jwks.py`, `key_origins.py`, `ip_pinned_transport.py`, `revocation_fetcher.py` and `canonical_formats/references.py`. `canonical.py`, the RFC 9421 signing path, is the one place that never calls it.

Consequence: a request to an IDN host signs and verifies over a signature base containing a raw non-ASCII host. A conformant counterparty computes the A-label base, so the two disagree and the signature fails to verify — the classic silent 9421 interop break.

## Failing cases

`canonicalization.json`:

| case | input | expected `@target-uri` | actual |
|---|---|---|---|
| `idn-to-punycode` | `https://bücher.example/p` | `https://xn--bcher-kva.example/p` | `https://bücher.example/p` |
| `idn-mixed-case-to-punycode` | `https://BÜCHER.Example/p` | `https://xn--bcher-kva.example/p` | `https://bücher.example/p` |

(`@authority` diverges identically: expected `xn--bcher-kva.example`, actual `bücher.example`.)

`negative/026-non-ascii-host.json` — the verifier side of the same gap:

- expected `request_signature_header_malformed`, actual `request_signature_invalid` at step 10.

The spec is explicit that these are two different obligations. Signers MUST emit A-labels; verifiers MUST **reject** a raw U-label rather than silently re-normalize it:

> A host containing raw non-ASCII bytes that has not been ToASCII-normalized by the signer MUST be rejected by the verifier; verifiers do not silently re-normalize.
> — `security.mdx`, `@target-uri` canonicalization step 2

## Processing mode matters

The spec pins the mode, because this is a known cross-language divergence:

[...]

Issue #976, opened 2026-07-28
Verifier does not reject malformed structured-field input at step 1 (wrong error code; covered-component smuggling)


KonstantinMirin:
Found while vendoring the complete AdCP 3.1.1 request-signing vector set (#975). Reproductions are the newly vendored vectors, currently strict-xfail in `tests/conformance/signing/test_verifier_vectors.py`.

## Problem

`verify_request_signature` has no step-1 strict-parse stage. Structured-field input that is malformed or ambiguous is not rejected at parse time; it flows on to a later checklist step and is rejected under a **different** error code — or, worse, is silently resolved to one interpretation.

Conformance requires byte-for-byte match on `expected_outcome.error_code`, so rejecting with the wrong code is a graded failure, not a near-miss. More importantly, each of these is a parser-differential surface: signer and verifier can disagree on what was actually signed.

## Failing vectors

| vector | expected | actual | step reached |
|---|---|---|---|
| `negative/021-duplicate-signature-input-label.json` | `request_signature_header_malformed` | `request_signature_components_incomplete` | 6 |
| `negative/022-multi-valued-content-type.json` | `request_signature_header_malformed` | `request_signature_invalid` | 10 |
| `negative/023-multi-valued-content-digest.json` | `request_signature_header_malformed` | `request_signature_invalid` | 10 |

### 021 — duplicate `Signature-Input` dictionary key

`Signature-Input` declares label `sig1` twice. RFC 8941 §3.2 permits an implementation to reject *or* retain the last value; the AdCP profile requires rejection at step 1 as downgrade protection.

This is the one with teeth. The SDK silently resolves the duplicate and then rejects for an unrelated reason — `request_signature_components_incomplete: required covered component missing: @authority` — which means it took the *second* `sig1`, the one covering fewer components. That is precisely the smuggling scenario the spec's rejection rule exists to prevent: a proxy appends a weaker `sig1` and a verifier that reads the last value grades a different component set than the signer committed to. The vector's own `$comment` calls this out.

### 022 — multi-valued `Content-Type`

`Content-Type: application/json, text/plain` on a field covered by the signature. `Content-Type` is not a List-Structured-Field, so concatenation leaves the covered value undefined relative to the signature base. The SDK derives a base from the raw value and fails at crypto verify instead of rejecting at parse.

### 023 — multi-valued `Content-Digest`

[...]

PR #980, opened 2026-07-28
test(signing): vendor the complete AdCP 3.1.8 request-signing vector set and pin it against drift

### LINKED ISSUE #975: Vendored request-signing conformance vectors are incomplete, and the loaders cannot detect it
## Summary

The vendored AdCP request-signing conformance vectors under `tests/conformance/vectors/request-signing/` are an incomplete copy of the spec set, and the tests that grade against them are structurally incapable of noticing.

Both graders load vectors by globbing and then assert only that the result is non-empty:

- `tests/conformance/signing/test_verifier_vectors.py:77-78` — `_load()` does `glob("*.json")` then `assert result, f"no vectors under ..."`
- `tests/conformance/signing/test_canonicalization.py:24-31` — same shape for vectors carrying `expected_signature_base`

Any non-empty subset satisfies that assertion. A missing vector is indistinguishable from a complete set, so the suite reports green on a partial copy — which is exactly what has been happening.

## The gap

Measured against AdCP **3.1.1**, the version `src/adcp/ADCP_VERSION` already targets (spec repo `dist/compliance/3.1.1/test-vectors/request-signing/`):

| | spec 3.1.1 | vendored | missing |
|---|---|---|---|
| `positive/` | 12 | 8 | `009-percent-encoded-unreserved-decoded`, `010-percent-encoded-slash-preserved`, `011-ipv6-authority`, `012-ipv6-authority-default-port-stripped` |
| `negative/` | 28 | 20 | `021-duplicate-signature-input-label`, `022-multi-valued-content-type`, `023-multi-valued-content-digest`, `024-unquoted-string-param`, `025-jwk-alg-crv-mismatch`, `026-non-ascii-host`, `027-webhook-registration-authentication-unsigned`, `028-unsigned-protocol-method-required` |
| `canonicalization.json` | ships | **absent** | the standalone signing-base fixture |

`canonicalization.json` is the conspicuous one: the vendored `README.md` documents it in the file-layout tree *and* devotes a full section to it, describing it as "the fastest way to surface cross-implementation divergence" and recommending SDKs "run `canonicalization.json` on every commit". The file was never copied in.

Three further vectors had silently drifted from their upstream contents: `016-replayed-nonce`, `017-key-revoked`, `020-rate-abuse`. `017` is the consequential one — it was vendored at a revision that revokes `test-ed25519-2026`, the primary signing key, rather than the dedicated `test-revoked-2026` keypair the spec now uses. `keys.json` had never been refreshed to carry that key either.

## Why it matters

Signature-base construction is the known-hard part of RFC 9421 interop, and the missing vectors sit exactly on its edges: `009`/`010` are percent-encoding, `011`/`012` are IPv6 authority, `026` is a non-ASCII host, `028` is the `protocol_methods_required_for` enforcement path.

A verifier can pass every currently-vendored vector and still reject validly-signed requests from a conformant counterparty. And because the SDK owns canonicalization on behalf of its consumers, these gaps are untested in every downstream implementation too, not just in this suite — this was found while adopting `adcp.signing` in a seller implementation that delegates canonicalization entirely to the SDK.

## Fix

PR #980 vendors the complete 3.1.1 set and, more importantly, replaces the non-empty assertion with a completeness pin so the set cannot silently drift again at the next spec bump. Adding the 12 files without that guard would only reset the clock.

The pin is over **content hashes**, not filenames — a filename-only manifest would not have caught the `016`/`017`/`020` drift. It is tied to `get_adcp_spec_version()`, so bumping the targeted spec version without re-vendoring fails loudly.

## Grading the new vectors turns up 12 real conformance failures

They are filed separately and marked strict-xfail in the PR rather than fixed there, since they are production behavior changes in a security-critical verifier:

- The verifier has no step-1 strict-parse stage, so malformed/ambiguous input flows through to a later check and is rejected under the wrong error code (negative `021`, `022`, `023`, `026`). Conformance requires byte-for-byte error codes, so "rejected anyway" is not a pass.
- `@target-uri` canonicalization applies no UTS-46 A-label conversion, drops a trailing empty query, and accepts five malformed authority shapes the spec requires be rejected.

KonstantinMirin:
Fixes #975.

## What was wrong

The vendored request-signing conformance vectors were an incomplete copy of the spec set, and the loaders could not tell. Both graders globbed the vector directory and asserted only that the result was non-empty (`test_verifier_vectors.py:77-78`, `test_canonicalization.py:24-31`), so any non-empty subset passed and a missing vector was indistinguishable from a complete set.

Against AdCP **3.1.8** — the version `src/adcp/ADCP_VERSION` already targets — the vendored copy was missing 4 of 12 positive vectors, 8 of 28 negative vectors, and `canonicalization.json` entirely. The vendored `README.md` documents `canonicalization.json` in its file-layout tree and devotes a section to it; the file had never been copied in.

Three further vectors (`016`, `017`, `020`) had silently drifted from upstream. `017` is the consequential one: it was vendored at a revision that revokes `test-ed25519-2026`, the primary signing key, rather than the dedicated `test-revoked-2026` keypair the spec now uses — and `keys.json` had never been refreshed to carry that key.

## What this PR does

**Vendors the complete set.** All 12 missing files plus `canonicalization.json`, and re-syncs the three drifted vectors, `keys.json` and `README.md`. The tree is now byte-identical to `dist/compliance/3.1.8/test-vectors/request-signing/` at spec tag `v3.1.8` — verified file-by-file in both directions.

**Replaces the non-empty assertion with a completeness pin.** This is the actual fix; adding 12 files without it would only reset the clock.

- `tests/conformance/signing/vectors.py` — one shared loader, replacing the two per-module glob helpers. Parametrization is gated on the pin, so a missing vector fails at **collection** time rather than quietly shrinking the graded set.
- `vector_manifest.json` — SHA-256 of every vendored file. Fails on a missing file, an unexpected extra file, **and** on content drift. The pin is over content rather than filenames precisely because a filename-only manifest would not have caught what happened to `016`/`017`/`020`. Regenerate with `python -m tests.conformance.signing.vectors --write`.
- `test_vector_completeness.py` — grades the pin, and ties it to `get_adcp_spec_version()` so bumping the targeted spec version without re-vendoring fails loudly instead of leaving the SDK graded against a stale contract. Also checks per-directory counts, contiguous vector numbering, and that every `jwks_ref` kid resolves in `keys.json` (that last one is what the stale `017` would have tripped).

**Wires in `canonicalization.json`.** It earns its own grader rather than being redundant with the per-vector `expected_signature_base` path: it exercises `canonicalize_target_uri` / `canonicalize_authority` over 31 cases, including 6 malformed-authority rejections that no signed vector covers at all. The spec's README calls it "the fastest way to surface cross-implementation divergence".

## The new vectors surface 12 real conformance failures

As expected, grading the previously-ungraded vectors turns the suite red. Nothing was weakened to get green — the failures are marked **strict**-xfail with the issue each one tracks, so a fix XPASSes and forces the entry to be retired rather than leaving a stale exemption behind.

They are filed separately rather than fixed here, since each is a production behavior change in a security-critical verifier and several change error codes that consumers may key on:

- **#976** — the verifier has no step-1 strict-parse stage, so malformed/ambiguous structured-field input flows through to a later check and is rejected under the wrong error code (`021`, `022`, `023`). `021` is the sharp one: the SDK silently resolves a duplicate `sig1` label to the *second* one, covering fewer components — exactly the smuggling case the spec's rejection rule exists to prevent.
- **#977** — `@target-uri` canonicalization applies no UTS-46 A-label conversion, and the verifier does not reject raw U-labels (`026` + 2 canonicalization cases). Note the SDK already ships `canonicalize_host()` and wires it into `jwks.py`, `key_origins.py`, `ip_pinned_transport.py` and `revocation_fetcher.py` — `canonical.py` is the one place that never calls it.
- **#978** — canonicalization accepts five malformed authority shapes the spec requires be rejected, and `request_target_uri_malformed` is not implemented at all (5 cases). Three of them collapse to an empty authority, so two different malformed inputs canonicalize to an identical signature base.
- **#979** — a trailing empty query is dropped, so `/p?` and `/p` produce the same signature base (1 case).

## Verification

- Vendored tree confirmed byte-identical to spec `v3.1.8`, both directions (nothing missing, nothing extra).
- Signing conformance suite: **459 passed, 9 skipped, 12 xfailed**.
- Full suite: **5911 passed, 41 skipped, 13 xfailed, 0 failed**.
- `ruff check src/` and `black` clean; `mypy` on the changed files introduces no new errors (the 3 remaining bare-`dict` warnings in `test_verifier_vectors.py` pre-date this branch, and `make typecheck` covers `src/adcp/` only).

**The completeness check was mutation-verified** — a guard that cannot fail is the same defect in a new costume, and this issue exists because an assertion that could not fail was trusted. Each mutation was applied, confirmed red, and reverted:

| mutation | result |
|---|---|
| delete `negative/026-non-ascii-host.json` | collection error, naming the missing file |
| delete `canonicalization.json` | collection error |
| edit a vector in place (content drift) | 1 failed — `vendored vectors differ from their pinned content` |
| add an unexpected `negative/029-not-in-spec.json` | collection error |
| add a ledger entry for a **passing** vector | 1 failed (strict xfail → XPASS), confirming the ledger self-retires |

No production code is touched; this change is vectors plus the guard that keeps them complete.

## Context

Found while adopting `adcp.signing` in a seller implementation that delegates canonicalization entirely to the SDK. Because the SDK owns canonicalization on behalf of its consumers, these gaps were untested in every downstream implementation too, not just in this suite.


KonstantinMirin:
Thanks for the review — addressed the `_write_manifest` nit in ac1e9a2a with a comment spelling out that the shallow copy must *replace* `files` rather than mutate it, since `_MANIFEST["files"]` is what `VECTOR_MANIFEST` is bound to and mutating it in place would rewrite the pin the tests grade against.

Also dropped a private `_pytest.mark.ParameterSet` import I'd used for the parametrization type hints (f0622470) — it was the only `_pytest` import in the repo and bought nothing, since `make typecheck` covers `src/adcp/` only.

Re-verified after both: signing conformance suite 459 passed / 12 xfailed, full suite 5911 passed / 0 failed, and `python -m tests.conformance.signing.vectors --write` regenerates the manifest byte-identically.

Agreed on prioritising #976 — the `021` duplicate-`sig1` case is the one with real teeth. It isn't just a code mismatch: the verifier resolves the duplicate to the *second* label, which covers fewer components, so the observable failure is `request_signature_components_incomplete: required covered component missing: @authority`. That's the smuggling path working end to end, and the wrong-code symptom is what makes it easy to misread as a benign parse difference.

KonstantinMirin:
CI note: 19/20 checks pass. The one red check, `AdCP storyboard runner — v3 reference seller (translator)`, is **not caused by this PR** — it is broken on `main`. Filed as #981.

I checked rather than assumed, because it did fail three times running here and other open PRs show green on that job:

- Those green results are stale — #974 last ran 2026-07-27, #970 on 07-07, #969 on 07-03. This branch was the only thing to run CI on 07-28, so there was no same-day comparison.
- Control: pushed `main` (`a79e29a2`) **unmodified** to a fork and ran the same workflow — https://github.com/KonstantinMirin/adcp-client-python/actions/runs/30398113683. The translator job fails there too, with the identical `Seller failed to start within 30s`, and it is the only failing job in that run.
- Not npm drift on the pinned alias: `@adcp/sdk` `latest` is still `12.1.1` from 07-26; the `13.0.0-rc.*` builds are on the separate `rc` tag.

So the boot regression landed between 07-27 19:24Z (last green) and 07-28 20:35Z, in a window with no CI runs. Details and a suggestion for making the step self-diagnosing (it currently discards the seller's output on timeout, which is why the log shows no cause) are in #981.

Everything this PR actually exercises is green: all four Python versions, Postgres conformance, schema validation, downstream import smoke, and the other five storyboard runners.

bokelley (changes requested):
Current main now targets AdCP 3.1.8 (#992), but this branch pins VECTOR_SET_SPEC_VERSION to 3.1.1. I merged the full stack into current main and ran tests/conformance/signing: 482 passed, 1 failed at test_manifest_pins_the_spec_version_the_sdk_targets (3.1.8 != 3.1.1).\n\nPlease rebase onto current main and update vector_manifest.json's spec_version/source to 3.1.8. I compared dist/compliance/3.1.1/test-vectors/request-signing with 3.1.8 at tag v3.1.8; the trees are byte-identical, so the file hashes/payload do not need to change.

KonstantinMirin:
Rebased onto current main and re-pinned at 3.1.8 in `d719b5c6`.

Only `spec_version` and `source` changed — regenerating the manifest reproduces all 43 file hashes identically, which independently confirms your byte-identical check of the two vector trees.

Worth noting this is the pin doing exactly what it was added for. It didn't detect a content drift; it forced an explicit re-vendor decision on a spec bump instead of letting the old vector set ride silently under a new version number. That silent-carry is the failure mode #975 was filed about.

One trap for the next bump, which cost me a cycle: `python -m tests.conformance.signing.vectors --write` regenerates `files` only and carries `spec_version`/`source` forward unchanged, so running it alone will never bump the version. Those two fields need editing first, then the regen confirms the hashes. The file's "do not hand-edit" banner reads as covering the whole document — I've noted the distinction in the commit message.

The three dependent PRs (#985, #987, #993) are restacked on this base and gated green.

`make ci-local`: 5994 passed, 0 failed.

bokelley (approved):
Reviewed the rebased 3.1.8 pin. The manifest version/source now match the SDK target, the vector payload hashes remain unchanged as expected, and CI is green. My previous request is fully addressed.

PR #983, opened 2026-07-28
ci: install Ladon PR reviewer, replacing the Argus workflow

andybevan-scope3:
## Why

Ladon is the shared AdCP PR reviewer, maintained in [`adcontextprotocol/actions`](https://github.com/adcontextprotocol/actions/tree/main/ladon) and consumed by floating major tag. This repo carried its own reviewer instead — a `claude-code-action` invocation driving a 245-line vendored prompt at `.github/ai-review/expert-adcp-reviewer.md` — so every engine fix had to be re-ported here by hand. Installed per [`ladon/INSTALL.md`](https://github.com/adcontextprotocol/actions/blob/main/ladon/INSTALL.md), following the "replacing an existing reviewer" path.

## What changed

- **`.github/workflows/ai-review.yml`** — replaced with the canonical consumer workflow from `INSTALL.md`, **copied verbatim**. Pins `adcontextprotocol/actions/ladon/review@ladon/review/v1` (tag resolves to `2b4b3aa`), keeps `model: claude-opus-4-8`, and preserves the `pull_request_target` posture: base-SHA-only checkout, PR head never fetched or executed, every `github.event.*` value reaching a shell via `env:`. Adds the workflow-modification gate the old workflow lacked.
- **Deleted `.github/ai-review/expert-adcp-reviewer.md`** — voice, severity model, largest-file rule, test-plan honesty, operational-readiness audit, and inline-comment format now come from the baseline rules in the actions repo (`ladon/reviewer/rules/*.md`). Nothing else referenced this file.
- **Added `LADON.md`** — the repo-specific half of the retired prompt, restated as parsed config. `## Repo Context` carries the mandatory rules that are only true here: semver signal on the `adcp.*` surface (release-please cuts a minor from `feat:`, so a breaking diff under it ships without a major), discriminated-union forward compat, the `types` import-layering allowlist, generated-code-is-not-source including the codegen renumbering-churn rule from `CLAUDE.md`, the `ctx_metadata` credential prohibition, and CI-gate regressions. Plus `## High-Risk Paths`, `## Trivial Paths`, `## Escalation Reviewers`, and `## Release Stack Branches`.

Job name `code_review` is unchanged, so the required-status-check context is preserved.

## Secrets

The reviewer moves from `IPR_APP_ID` / `IPR_APP_PRIVATE_KEY` to `SECRETARIAT_APP_ID` / `SECRETARIAT_APP_PRIVATE_KEY`. All of `SECRETARIAT_APP_ID`, `SECRETARIAT_APP_PRIVATE_KEY`, and `ANTHROPIC_API_KEY` already resolve to this repo as org-level secrets (verified via `repos/.../actions/organization-secrets`) — no new secrets to add.

**One prerequisite I could not verify:** whether the **AAO Secretariat** GitHub App is installed on this repo (reading that requires org-admin scope). Per `INSTALL.md`, secrets existing is not the same as the App being installed — if the token-mint step fails on the first run *without* a secret-not-found error, that is the cause.

## Notes for the reviewer

Two deltas where verbatim-copy conflicts with what this repo was already doing. `INSTALL.md` says "copy it verbatim: do not reconstruct it," so I did, but both are worth a decision:

1. **Action pins go down.** The template uses `actions/checkout@v5` and `actions/create-github-app-token@v1`; the Argus workflow used `@v6` and `@v3`. Same as `adcp-client`'s live workflow.
2. **`persist-credentials: false` is gone** from the checkout step — the Argus workflow set it, the template does not. Low impact (nothing pushes, and PR-head code is never executed), but it was real hardening. Happy to add it back if you'd rather keep it.

Also: `ladon/review`'s `app-id` input is marked deprecated in favour of `client-id`, but `INSTALL.md`'s template still uses `app-id` and the action still accepts it. Left as-is to stay on the documented template.

No `## Gated Paths` section — this repo has no `CODEOWNERS` (removed in #968), and per `AUTHORING.md` a gated path without a matching CODEOWNERS + branch-protection rule becomes a permanently un-liftable block.

I dropped a `## Skip Bot Authors` section I had initially written: the action's own default is `dependabot[bot],renovate[bot],github-actions[bot]`, and because LADON.md *overrides* rather than extends that default, listing only `dependabot[bot]` would have narrowed coverage. `## Release Stack Branches` handles the release-please PRs (branch `release-please--branches--main--components--adcp`, authored by `app/aao-ipr-bot`) deterministically instead.

## Test plan

- [x] Workflow YAML parses; single job `code_review`; trigger, `paths-ignore`, permissions, and all five steps verified.
- [x] `LADON.md` verified against the real parser (`ladon/setup/src/ladon-md.ts`) — all six H2 headings resolve, 17 high-risk globs / 3 trivial globs / 1 reviewer / 1 release branch parse, `## Repo Context` captures all 7 H3 rule blocks.
- [x] `ladon/review/v1` tag exists; passed inputs match the action's declared inputs at that tag.
- [x] `SECRETARIAT_APP_ID` / `SECRETARIAT_APP_PRIVATE_KEY` / `ANTHROPIC_API_KEY` confirmed available to this repo.
- [x] No remaining references to `.github/ai-review/` anywhere in the tree.
- [x] PR title passes `scripts/check-pr-title.cjs`.
- [ ] **End-to-end run — not exercisable from this PR.** `pull_request_target` registers from the default branch and runs from the base SHA, and the workflow-mod gate makes Ladon decline to self-review. This PR needs a human approver; Ladon starts on the next PR. After merge, confirm registration with:
  ```
  gh api "repos/adcontextprotocol/adcp-client-python/actions/workflows" --jq '.workflows[] | "\(.path) \(.state)"'
  ```
  `.github/workflows/ai-review.yml` must be listed and `active`.


benminer (approved):

bokelley (dismissed):
Refreshed onto v7.0.1-era main and revalidated the canonical Ladon consumer workflow, base-SHA-only security posture, preserved code_review status context, Opus 4.8 model pin, and LADON.md parser structure. Approving the reviewer migration.

bokelley (dismissed):
Re-approved after the focused workflow-security fix: GitHub primitives are SHA-pinned, the gate token is limited to pull-request writes, the intentional Ladon v1 float is documented and narrowly exempted, and local actionlint plus zizmor 1.29.0 report clean.

bokelley (approved):
Final re-approval: Ladon is now pinned to the reviewed v1 commit because it receives the Secretariat private key, checkout credentials are not persisted, all GitHub primitives are SHA-pinned, App-token scope is least-privilege, and actionlint plus zizmor 1.29.0 are clean.

PR #985, opened 2026-07-29
fix(signing): reject malformed authorities, convert IDN hosts to A-labels

### LINKED ISSUE #977: @target-uri canonicalization applies no IDNA A-label conversion, and the verifier does not reject raw U-labels

### LINKED ISSUE #978: Canonicalization accepts five malformed authority shapes the spec requires be rejected; request_target_uri_malformed unimplemented

KonstantinMirin:
> **Stacked on #980 — do not merge before it.**
> Own commit: `9be012b8`. Everything else in the diff below is #980's vendored vectors.
> Incremental diff: https://github.com/KonstantinMirin/adcp-client-python/compare/fix/vendor-complete-signing-vectors...fix/target-uri-malformed-authority

Fixes #978. Partially fixes #977 (signer half — see *Why #977 is only half-closed* below).

## What was wrong

Six authority shapes the profile requires be rejected were canonicalized instead, and `request_target_uri_malformed` was not implemented at all. Separately, the signing path never called the package's own host canonicalizer, so an IDN authority signed and verified under a raw U-label.

## What changed

**The rejection set lives inside `_canon_authority`**, so `@target-uri` and `@authority` inherit it once rather than each carrying a copy. `urlsplit`'s own bare `ValueError` is normalized onto the same typed error — without that, `malformed-ipv6-missing-closing-bracket` passes on someone else's exception and grades nothing. (#980 had exactly this false green; it was fixed there so all six reject cases xfail together, and they retire together here.)

**The predicate is ported minus one branch.** The reference implementation rejects raw non-ASCII hosts. That is correct *there*, because it wraps a verifier — but `canonical.py` is shared by signer and verifier, and the signer is required to **convert** a U-label to its A-label form (`idn-to-punycode`), not refuse it. Porting that branch verbatim would have made the two halves of #977 fight each other. The test survives as `host_has_raw_non_ascii`, which PR B's step-1 precheck imports rather than re-deriving: **one rule, two codes, two steps.**

**Host canonicalization reuses `_idna_canonicalize.canonicalize_host`** instead of hand-rolling IDNA here. That helper already has six consumers and `canonical.py` was conspicuously absent from them — adding a private IDNA step would have made this the seventh host normalizer in the tree.

The ASCII fast path is load-bearing, not an optimization: the helper strips IPv6 brackets and raises on underscore hosts and >63-byte labels, all of which sign correctly today.

**Failure mode is closed**, deliberately. The existing callers are split — `jwks.py` and `revocation_fetcher.py` fail closed, `ip_pinned_transport.py` and `key_origins.py` fail open — so precedent was not available to appeal to. The reason to fail closed here is that a signature base must never be computed over a host we could not canonicalize; the fail-open callers are comparison paths, where raising would turn a mismatch into an outage.

## Behaviour changes — both wire-visible

**1. A trailing FQDN-root dot is now stripped.** `https://example.com./p` derives authority `example.com`, where it derived `example.com.` before. A signer on this SDK and a verifier on an older one will disagree for FQDN-root URLs.

This one deserves scrutiny, because the spec does not decide it. `canonicalize_host` strips the root dot as part of UTS-46 preparation; a bare `.lower()` keeps it. Branching first would therefore make `example.com.` and `bücher.example.` normalize *differently* — a signer emitting the A-label form and a verifier reading a Host-header U-label would derive two different authorities for the same host. Stripping once, before the branch, is what makes the two agree, and it makes them agree with the package's designated normalizer rather than with an accident of the old code.

**No vector covers this.** I could not add one: `canonicalization.json` is SHA-256 pinned in `vector_manifest.json`, and `test_vendored_tree_matches_manifest_exactly` fails on extra files as well as changed ones — correctly, since #980 exists to stop exactly that drift. So the behaviour is pinned by an SDK-local test and tracked in #986, which asks that the case be defined upstream and shipped as a real vector. When that lands, the local test is superseded.

**2. Authority-less URIs now raise.** `canonicalize_target_uri("mailto:x@example.com")` and relative paths round-tripped unchanged before; both are public exports.

## The webhook code (resolved — this section previously said the opposite)

`request_target_uri_malformed` retags onto `webhook_target_uri_malformed`, which the spec defines: security.mdx's webhook checklist lists it in the error taxonomy (L1469, L1566) and requires it at step 10 for a malformed or mismatched authority. The name deliberately drops the `webhook_signature_` prefix the rest of the family carries, because that is how the spec names structural codes.

An earlier revision of this PR mapped onto `webhook_signature_header_malformed` and this section argued for it on the grounds that "no spec text defines a webhook twin". That was wrong — it reflected a grep of this repo rather than the spec. Corrected in `7548cc84`; the paragraph is kept here only so the reasoning trail is visible rather than silently rewritten.

## Why #977 is only half-closed

#977 has two halves. The signer half — no A-label conversion — is fixed here and graded by the two positive `idn-*` vectors. The verifier half — a raw U-label arriving on the wire must be *rejected*, not converted — is graded by `negative/026`, which expects `request_signature_header_malformed` at **step 1**. That is #976's precheck, so it lands in PR B and closes #977 there. `026`'s ledger entry is retagged accordingly rather than left naming an issue whose other clause is now false.

Worth stating plainly: between this PR and PR B, a verifier will silently **convert** a wire U-label where the spec says it must refuse. That is an argument for landing PR B promptly rather than letting it sit.

## Notes

- **`webhook_transport_hooks.py:96-100`** synthesizes a netloc from the unvalidated `rewrite_to` config string. An IPv6 value (`rewrite_to="::1"`) yields an unbracketed netloc that `_canon_authority` mis-splits at the last colon *today*; after this PR it becomes a hard rejection. Failing loudly beats mis-signing, but anyone with that config will notice.
- `https://host:/p` still escapes as an uncoded `ValueError` from `int("")`, so "every authority rejection carries `.code`" is not literally true yet. Left alone deliberately — a gate that cannot *prove* malformation should pass traffic through.
- Titled `feat(signing):` so release-please cuts a **minor**. Both behaviour changes above are adopter-visible, and a rejection-set expansion arriving in a patch is how integrators get surprised. Retitle to `fix:` if you'd rather it went out as a patch.

## Test plan

`make ci-local`: **5953 passed, 0 failed, 41 skipped, 6 xfailed** (baseline on #980's branch: 5938 passed / 14 xfailed). The delta is exactly +15 passed / −8 xfailed: eight retired ledger entries plus seven new tests.

Two gaps the vectors provably do not cover, now covered — a probe across the full suite found **no** test driving either path:

- `test_target_uri_malformed_boundary.py` drives `verify_request_signature` and `verify_webhook_signature` with a gate-tripping URL. The verifier one guards clause *ordering*: `TargetUriMalformedError` is a `ValueError`, so moving the new `except` below the existing `except (ValueError, KeyError)` silently downgrades every rejection to `request_signature_header_malformed`. Without this test that reorder is invisible.
- `test_canonicalization.py` pins root-dot symmetry across both host branches.

### DIFF
--- a/src/adcp/signing/canonical.py
+++ b/src/adcp/signing/canonical.py
@@ -267,6 +267,14 @@ def _canon_host(host: str, netloc: str) -> str:
     """
     if host.endswith("."):
         host = host[:-1]
+    if not host or any(label == "" for label in host.split(".")):
+        # Re-checked AFTER the strip. `_malformed_authority_reason` runs on the
+        # raw netloc, where `.` and `..` are non-empty and look like hosts; it
+        # is only stripping the root dot that empties them. Without this,
+        # `https://./p` canonicalized to `https:///p` -- the empty authority
+        # this very module rejects two functions up, reached by a path that
+        # skipped the check.
+        raise TargetUriMalformedError(netloc, "the authority carries no host once normalized")
     if not host_has_raw_non_ascii(host):
         return host.lower()
     try:
--- a/src/adcp/signing/canonical.py
+++ b/src/adcp/signing/canonical.py
@@ -236,16 +236,49 @@ def _canon_authority(netloc: str, scheme: str) -> str:
         host = netloc[: end + 1]
         tail = netloc[end + 1 :]
         if tail.startswith(":"):
-            port = int(tail[1:])
+            port = _port_or_reject(tail[1:], netloc)
     elif ":" in netloc:
         host, portstr = netloc.rsplit(":", 1)
-        port = int(portstr)
+        port = _port_or_reject(portstr, netloc)
     else:
         host = netloc
     host = _canon_host(host, netloc)
     if port is not None and port != _DEFAULT_PORTS.get(scheme):
         return f"{host}:{port}"
     return host
+
+
+_ASCII_DIGITS = frozenset("0123456789")
+
+
+def _port_or_reject(portstr: str, netloc: str) -> int | None:
+    """Parse a port per RFC 3986 §3.2.3 (`port = *DIGIT`), or reject.
+
+    The port used to go straight into `int()`, which is far more permissive
+    than the grammar and produced three distinct problems:
+
+    * `int("-80")` gave the authority `host:-80`, which is not an authority.
+    * `int("8_0")` is 80 -- Python accepts underscore digit separators.
+    * `int("٨٠")` is also 80 -- `int()` accepts non-ASCII digits, so
+      `host:٨٠` and `host:80` collapsed to the SAME canonical authority. A
+      peer that does not fold Arabic-Indic digits derives a different
+      `@authority` from identical bytes, and the signature fails for a reason
+      neither side can see in its own logs.
+
+    `str.isdigit()` does not close that last one -- `"٨٠".isdigit()` is True --
+    so the test is ASCII digits specifically.
+
+    An EMPTY port is legal and means "default": the grammar is `*DIGIT`, and
+    §3.2.3 says a normalizer should drop the port and its colon when empty. So
+    `https://host:/p` normalizes to `host` rather than being rejected.
+    """
+    if not portstr:
+        return None
+    if not all(ch in _ASCII_DIGITS for ch in portstr):
+        raise TargetUriMalformedError(
+            netloc, f"the port {portstr!r} is not a sequence of ASCII digits"
+        )
+    return int(portstr)
 
 
 def _canon_host(host: str, netloc: str) -> str:
--- a/src/adcp/signing/canonical.py
+++ b/src/adcp/signing/canonical.py
@@ -12,7 +12,45 @@ from __future__ import annotations
 
 from collections.abc import Mapping
 from dataclasses import dataclass
-from urllib.parse import urlsplit, urlunsplit
+from urllib.parse import SplitResult, urlsplit, urlunsplit
+
+import idna
+
+from adcp.signing._idna_canonicalize import canonicalize_host
+
+#: Spec code for an authority the AdCP profile requires be rejected rather than
+#: canonicalized. Declared here rather than imported from ``errors`` so this
+#: module stays a leaf: ``errors`` may depend on canonicalization, never the
+#: reverse. ``verifier`` maps this onto the wire error at its boundary.
+REQUEST_TARGET_URI_MALFORMED = "request_target_uri_malformed"
+
+
+class TargetUriMalformedError(ValueError):
+    """A URL whose authority the profile requires be rejected.
+
+    Subclasses ``ValueError`` because every existing caller of the
+    canonicalizers already treats a bad URL as a ``ValueError``; carrying
+    ``.code`` lets the verifier boundary emit the spec code instead of
+    re-deriving one. The ``reason`` names which rule fired, so a rejection
+    message says what was wrong rather than restating "malformed".
+    """
+
+    code = REQUEST_TARGET_URI_MALFORMED
+
+    def __init__(self, subject: str, reason: str) -> None:
+        super().__init__(f"{reason}: {subject!r}")
+        self.reason = reason
+
+
+def host_has_raw_non_ascii(host: str) -> bool:
+    """Whether *host* carries raw non-ASCII bytes (an un-normalized U-label).
+
+    One definition, deliberately two call sites with two different codes: here
+    it selects the UTS-46 branch on the signing path, while the verifier's
+    header precheck uses it to reject a U-label arriving on the wire as
+    ``request_signature_header_malformed``. Share the predicate, never the code.
+    """
+    return not host.isascii()
 
 
 @dataclass(frozen=True)
@@ -80,7 +118,7 @@ def build_signature_base(
 
 def canonicalize_target_uri(url: str) -> str:
     """Produce the `@target-uri` derived-component value per AdCP profile."""
-    parts = urlsplit(url)
+    parts = _split_or_reject(url)
     scheme = parts.scheme.lower()
     netloc = _canon_authority(parts.netloc, scheme)
     path = _normalize_path(parts.path)
@@ -93,8 +131,23 @@ def canonicalize_target_uri(url: str) -> str:
 
 def canonicalize_authority(url: str) -> str:
     """Produce the `@authority` derived-component value per AdCP profile."""
-    parts = urlsplit(url)
+    parts = _split_or_reject(url)
     return _canon_authority(parts.netloc, parts.scheme.lower())
+
+
+def _split_or_reject(url: str) -> SplitResult:
+    """`urlsplit`, with its bare refusals normalized onto the typed error.
+
+    `urlsplit` rejects some malformed authorities itself -- `https://[::1/p`
+    among them -- with a plain `ValueError` carrying no code. That refusal is
+    correct but anonymous, so it is re-raised here as the same typed error the
+    authority gate raises. Without this, one of the six malformed-authority
+    vectors would "pass" on someone else's exception.
+    """
+    try:
+        return urlsplit(url)
+    except ValueError as exc:
+        raise TargetUriMalformedError(url, f"the URL could not be parsed ({exc})") from exc
 
 
 _DEFAULT_PORTS = {"http": 80, "https": 443}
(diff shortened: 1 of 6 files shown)


bokelley (changes requested):
Two corrections before merge:\n\n1. The spec does define the webhook twin. AdCP 3.1.1's webhook checklist names webhook_target_uri_malformed as a structural code, step 10 requires it for malformed/mismatched authority, and the webhook error taxonomy lists it explicitly: https://github.com/adcontextprotocol/adcp/blob/v3.1.1/docs/building/by-layer/L1/security.mdx#L1469-L1483 and https://github.com/adcontextprotocol/adcp/blob/v3.1.1/docs/building/by-layer/L1/security.mdx#L1549-L1567. Please add WEBHOOK_TARGET_URI_MALFORMED and map REQUEST_TARGET_URI_MALFORMED to it; mapping to webhook_signature_header_malformed emits the wrong stable wire code.\n\n2. _canon_host strips a trailing root dot after _malformed_authority_reason has accepted a non-empty host. Consequently https://./p canonicalizes to an empty authority instead of rejecting. Please reject when the post-strip host is empty and add a regression test.\n\nThe rest of the malformed-authority/IDNA implementation looks sound.

KonstantinMirin:
Both fixed in `7548cc84`.

**1. The webhook twin.** You're right, and my claim was wrong for a specific reason worth naming: I grepped *this repo* for `webhook_*_malformed`, found nothing, and reported "the spec defines no webhook twin" as a finding. That was only the limit of where I looked, stated with far more confidence than the evidence supported.

Added `WEBHOOK_TARGET_URI_MALFORMED` and mapped `REQUEST_TARGET_URI_MALFORMED` onto it, with a comment noting the name deliberately breaks the `webhook_signature_` prefix the rest of the family carries because the spec names it that way. I also rewrote the retag test's docstring, which had my wrong reasoning baked in as justification and would have read to the next person as a considered spec finding.

**2. The empty authority.** Reproduced, and slightly worse than reported: `https://./p` was producing `https:///p` — exactly the `malformed-empty-authority` shape this PR rejects — reached by a path that ran the gate before the root-dot strip could empty the host. `https://../p` yielded the authority `.`.

`_malformed_authority_reason` judges the raw netloc, where `.` is non-empty and passes as a host, so the check has to run again after the strip. Stated as "no empty label" rather than "not empty", so `https://a..b/p` is covered too. Regression test over all four shapes; `example.com.`, `bücher.example.` and `[::1]` still canonicalize normally.

Also restacked onto the 3.1.8 base (see #980). `make ci-local`: 6013 passed, 0 failed.

KonstantinMirin:
Thanks — the port find is real and the stale-body catch was correct. Both addressed in `4331ac84`. One correction to the suggested remedy, and the semver point I'm deferring rather than deciding.

## Port validation — fixed, with a change to the prescription

Gating `portstr` on digits-only would reject `https://host:/p`, but RFC 3986 §3.2.3 defines `port = *DIGIT`, so an **empty port is legal**, means "default", and normalizers are directed to drop it along with its colon. `urlsplit` agrees (`port=None`). That is also the better answer to the bare-`ValueError` complaint: the correct outcome there was never an error at all, so it now normalizes to `host` rather than raising anything.

Second, and this one matters more: **`str.isdigit()` does not close the case it was aimed at.** `"٨٠".isdigit()` is `True` in Python, so a digits-only gate would leave `host:٨٠` and `host:80` folding to the same canonical authority — exactly the differential you flagged, surviving the fix meant to remove it. The gate tests ASCII digits explicitly, and both the code and the test docstring say why, so nobody "simplifies" it back to `isdigit()` later.

Resulting rules:

```
https://host:/p       -> host        (empty port omitted, per §3.2.3)
https://[::1]:/p      -> [::1]
https://host:8443/p   -> host:8443   (unchanged)
https://host:-80/p    -> request_target_uri_malformed
https://host:8_0/p    -> request_target_uri_malformed   (int() accepts _ separators)
https://host:٨٠/p     -> request_target_uri_malformed   (int() accepts non-ASCII digits)
https://host:8a/p     -> request_target_uri_malformed
```

Six tests red before the change. You're right that this was the same "passes on someone else's exception" failure `_split_or_reject` exists to prevent, one frame lower — it belonged in this PR rather than a follow-up, since the PR's own claim is that every authority rejection carries the spec code.

## Stale PR description — fixed

Correct, and worth stating plainly: the body argued for mapping onto `webhook_signature_header_malformed` on the grounds that no spec text defined a webhook twin. That claim came from grepping *this repo* rather than the spec, and `7548cc84` had already overturned it. The section is rewritten, with the wrong reasoning left visible rather than silently deleted so the trail is readable.

## Semver — deferring to @bokelley, who has already ruled

He retitled these `feat` → `fix`, on the grounds that they restore spec compliance and that adopter-visible behaviour alone doesn't make something a feature. I've aligned the commit subjects to match, since the PR title and the commit subject disagreeing would have made the release type depend on which merge button got pressed.

I think you and he are arguing different axes rather than disagreeing: his point is that this isn't a *feature*, yours is that it's *breaking*. Both can hold — a fix can be breaking — and `fix(signing)!:` is the construction that satisfies both readings if he wants the major.

For what it's worth as the author, I'd argue against the major: the only signed-byte change is the trailing FQDN-root dot, the spec is silent on it and ships no vector (#986 asks upstream to rule), and a major bump over an unspecified edge case seems disproportionate at this stage. But it's his release and his call — I'm not going to settle it by retitling, which is why the `!` isn't there.

## Follow-ups you raised

- **Root-dot needs an upstream ruling** — agreed, and that's exactly what #986 asks for. The SDK-local test is explicitly marked as superseded once a real vector exists.
- **The interim U-label window** — agreed; PR B (#993) is ready and gated, so the window closes as soon as the stack lands.

bokelley (dismissed):
Reviewed the latest head, including the spec-defined webhook_target_uri_malformed mapping, the post-normalization empty-host rejection, and RFC 3986 port validation. I am treating these as a patch-level spec-conformance correction rather than a major-version trigger. The stacked signing suite passes locally: 533 passed, 4 expected xfails.

Issue #777, opened 2026-05-21
Migrate IDNA canonicalization from stdlib (IDNA 2003) to PyPI `idna` (IDNA 2008) across signing/


bokelley:
## Context

ADCP #3690 security.mdx line 1132 mandates **IDNA-2008 A-label form** for host canonicalization on origin comparisons. The package currently uses **stdlib ``host.encode(\"idna\")``** (IDNA 2003) in four places:

- ``src/adcp/signing/jwks.py:201``
- ``src/adcp/signing/ip_pinned_transport.py:110``
- ``src/adcp/signing/revocation_fetcher.py:380``
- ``src/adcp/signing/key_origins.py`` (added in #775)

#775's commit body acknowledged this divergence and made the deliberate call to match the existing convention rather than fragment the canonicalization story across modules. This issue tracks the package-wide migration.

## Practical impact (today)

IDNA 2003 vs IDNA 2008 differ on a small set of code points:

- ``ß`` (German sharp s) — IDNA 2003 maps to ``ss``; IDNA 2008 preserves as ``xn--strae-oqa``.
- Final sigma ``ς`` — IDNA 2003 maps to ``σ``; IDNA 2008 preserves.
- ZWJ / ZWNJ — IDNA 2003 strips; IDNA 2008 may preserve in valid contexts.

For ad-tech operator domains this is near-zero risk in practice. The real exposure is **cross-SDK conformance**: a peer SDK using the ``idna`` package would canonicalize the same wire bytes differently, and the two would disagree on whether ``straße.de`` and ``xn--strae-oqa.de`` are the same origin.

## What to change

1. Add ``idna`` to ``pyproject.toml`` ``[project.dependencies]``.
2. Replace ``host.encode(\"idna\")`` with ``idna.encode(host, uts46=True)`` in all four callsites in a single commit.
3. Update the canonicalization rationale comments in each module (the four currently cross-reference each other; the migration should update all of them).
4. Add a regression test: ``straße.de`` canonicalizes to ``xn--strae-oqa.de`` (IDNA 2008 behavior), not ``strasse.de`` (IDNA 2003 behavior). Verifies the migration actually changed behavior.
5. Run the storyboard + conformance suites to confirm no regressions in the existing IDNA paths.

## Why one commit

Splitting the migration across PRs would leave the codebase in a state where two SDK modules canonicalize the same host differently. Keep all four callsites coherent — one commit, one review.

## References

[...]

PR #995, opened 2026-07-29
fix(signing): idna-normalize hosts in etld+1 binding

### LINKED ISSUE #988: eTLD+1 binding does not IDNA-normalize, so a U-label agent URL fails to bind to an A-label brand domain
## Summary

`adcp.signing.etld.host_from` applies no IDNA conversion, so `same_registrable_domain` — the eTLD+1 gate the brand-authorization binding runs on — returns `False` when the two sides spell the same IDN host differently. A brand that publishes its `brand.json` under a U-label host and lists its agent under the A-label form (or the reverse) fails the binding.

Separately, the same function is asymmetric with itself on the trailing FQDN-root dot, and its docstring describes behaviour that only one of its two branches has.

## Cause

`src/adcp/signing/etld.py:73-82`:

```python
if "://" in value:
    parts = urlsplit(value)
    host = parts.hostname
    if not host:
        raise ValueError(f"URL has no host: {value!r}")
    return host.lower()                              # :78 — no dot strip, no IDNA
stripped = value.strip().rstrip(".").lower()         # :79 — dot stripped, still no IDNA
```

Neither branch performs UTS-46 / IDNA-2008 encoding, and `tldextract` does not do it either — it accepts a U-label and returns a U-label eTLD+1. So the two spellings never reduce to a common form.

The docstring at `:63-64` claims the function "trims a single trailing dot (the FQDN root separator) so `Example.COM.` and `example.com` compare equal". The bare-host branch does that; the URL branch does not. The bare-host branch also uses `rstrip(".")`, which strips *all* trailing dots, not one.

## Reproduce

```python
from adcp.signing.etld import host_from, registrable_domain, same_registrable_domain

# IDNA gap — the load-bearing one
registrable_domain("straße.de")            # 'straße.de'
registrable_domain("xn--strae-oqa.de")     # 'xn--strae-oqa.de'
same_registrable_domain("https://shop.straße.de/", "xn--strae-oqa.de")
# False   <- same host, two spellings, binding fails

# Trailing-dot asymmetry
host_from("https://Example.COM./")   # 'example.com.'
host_from("Example.COM.")            # 'example.com'
host_from("https://x.example.com..") # 'x.example.com..'
host_from("x.example.com..")         # 'x.example.com'
```

## Impact

**The IDNA gap is a real binding failure, fail-closed.** `src/adcp/signing/brand_authz.py:255` calls `same_registrable_domain(agent_url, brand_host)` as step 2a of the eTLD+1 binding. For `brand_domain="bücher.example"` and a `brand.json` listing `https://ads.xn--bcher-kva.example/`, the agent is found by `_find_listed_agents` (byte-equal URL match) but step 2a returns `False`, the `authorized_operators[]` fallback does not match either, and the result is `BrandAuthorizationResult(False, reason="binding_failed")` (`brand_authz.py:278-283`). A legitimate agent is refused; nothing is wrongly *accepted*. Scope is limited to brands on IDN hosts, which is a small population — but the failure is silent from the publisher's side and `binding_failed` points at the wrong thing.

**The trailing-dot asymmetry does not currently affect the binding.** `tldextract` strips trailing dots itself, so `registrable_domain` and `same_registrable_domain` return identical results for both spellings — verified above (`registrable_domain("https://example.com.")` and `registrable_domain("example.com")` both give `example.com`). The defect is observable only through `host_from`, which is a public export (`src/adcp/signing/__init__.py:195,430`), and through the docstring, which is wrong. Worth fixing as correctness and as a precondition for the IDNA change; not worth describing as a security issue.

## The workaround already in the tree

`src/adcp/adagents.py:378-383` does exactly the conversion `host_from` is missing, by hand, at its own call site:

```python
def _idna_ascii_host(hostname: str, context: str) -> str:
    try:
        return idna.encode(hostname.rstrip("."), uts46=True).decode("ascii").lower()
    except idna.IDNAError as e:
        raise AdagentsValidationError(...) from e
```

and applies it to both arguments before calling `same_registrable_domain` (`adagents.py:402-404`). That is the redirect path working around this gap locally. It is also a hand-retyped copy of `_idna_canonicalize.canonicalize_host` steps 1/2/4 and diverges from it in three ways: no leading/trailing `.strip()`, `transitional=` left to the `idna>=3` default rather than pinned the way the helper pins it, and no IP-literal short-circuit (so a public IPv6-literal redirect target raises "invalid IDNA hostname" rather than passing through). None of the three is reachable today — `_normalize_domain` and `_check_safe_host` run upstream of both call sites — so this is a maintenance duplicate rather than a live bug. It disappears entirely if `host_from` does the conversion.

## Suggested direction

Route both branches of `host_from` through `_idna_canonicalize.canonicalize_host`, which already strips exactly one trailing dot, lowercases, short-circuits IP literals, and encodes with `uts46=True, transitional_processing=False`. `canonical_formats/references.py:38` establishes that a non-`signing` module may import that helper, so `adagents.py` can drop `_idna_ascii_host` in the same change.

Two decisions the change has to make explicitly rather than incidentally:

1. **`registrable_domain()` return values change for IDN inputs.** It is public API; today `straße.de` returns `straße.de`, after the change `xn--strae-oqa.de`. Both sides of `same_registrable_domain` move together so the predicate stays correct, but adopters logging or persisting the value see it change.
2. **`canonicalize_host` raises** on underscore hosts and labels over 63 bytes, where `host_from` returns them unchanged today. The existing six callers are split between fail-open (`ip_pinned_transport.py:119`, `key_origins.py`) and fail-closed (`jwks.py`, `revocation_fetcher.py`), so "match the convention" is not available as an answer. Given `etld` is a comparison path and `registrable_domain` already documents a fail-closed `None` contract, mapping the raise onto `None` is probably right — but it should be stated in the diff.

## Related

- #977 / #985 — the same missing A-label conversion on `canonical.py`'s signature-base path. Different module, different consumer (`@authority` vs the brand binding); fixing one does not fix the other.
- #986 — trailing-root-dot handling in `canonical.py`. That issue asks upstream to define the rule for the *signature base*, where the byte value is on the wire. `etld`'s dot handling is internal to eTLD+1 comparison and is not spec-visible, so it does not need the same upstream decision.
- #777 (closed) — migrated IDNA canonicalization from the stdlib (IDNA 2003) to the PyPI `idna` package (IDNA 2008) "across `signing/`". `etld.py` was not converted in that pass; `host_from` still does a bare `.lower()`. This is residue of that migration rather than a new regression — which is also why the fix here is to call the helper #777 established rather than to introduce another one.

KonstantinMirin:
Fixes #988.

Independent of the signing stack (#980/#985/#987/#993) — branched off `main`, touches no file any of them touch.

## What was wrong

`etld.host_from` applied no IDNA conversion, so `same_registrable_domain` — the eTLD+1 gate the brand-authorization binding runs on — returned `False` when the two sides spelled the same IDN host differently. A brand publishing `brand.json` under a U-label host and listing its agent under the A-label form (or the reverse) failed the binding.

It was also asymmetric with itself: the URL branch never stripped the trailing FQDN dot despite the docstring claiming it did, so `host_from("https://Example.COM./")` → `"example.com."` while `host_from("Example.COM.")` → `"example.com"`.

Measured before / after:

```
registrable_domain("straße.de")                                  "straße.de"   -> "xn--strae-oqa.de"
same_registrable_domain("https://shop.straße.de/", "xn--strae-oqa.de")  False  -> True
host_from("https://Example.COM./")                               "example.com." -> "example.com"
```

## How

Both branches of `host_from` now delegate to `_idna_canonicalize.canonicalize_host` — the helper #777 established and that `jwks`, `revocation_fetcher`, `key_origins` and `ip_pinned_transport` already use. This finishes that migration rather than forking it; `etld.py` was simply missed in the original pass. **No new normalizer**: the change deletes hand-rolled logic rather than adding a seventh implementation to a tree that already had six.

## The fail-open vs fail-closed decision

`canonicalize_host` raises on underscore hosts, labels over 63 bytes, and leading hyphens, where `host_from` previously passed them through. The existing callers are **split** — `jwks` and `revocation_fetcher` fail closed, `ip_pinned_transport` and `key_origins` fail open — so "match the convention" was not available as an answer.

This fails **closed**: `registrable_domain` already documents a `None` contract for unusable input, and a binding gate that cannot canonicalize a host should refuse to bind rather than guess. `idna.IDNAError` subclasses `UnicodeError` subclasses `ValueError`, so `brand_authz.py:207`'s existing `except ValueError` still catches it — an IDNA-invalid `brand_domain` degrades to `reason="brand_domain_invalid"` instead of escaping mid-verification. The reasoning is in the module docstring and the commit message, as #988 asked.

## Adopter-visible changes

- `registrable_domain` returns **A-labels** for IDN input (`xn--strae-oqa.de`, not `straße.de`).
- `host_from` now raises for hosts IDNA cannot process; `registrable_domain` maps that to `None`.
- UTS-46 folding means compatibility forms now compare equal — `same_registrable_domain("ｅｘａｍｐｌｅ.ｃｏｍ", "example.com")` is now `True`. Correct IDNA behaviour and consistent with how DNS resolves it, but it is a widening beyond the U-label/A-label case the issue described, so calling it out explicitly.

## Test plan

`tests/conformance/signing/test_etld.py`, new. **7 of its 10 tests fail** against the pre-fix module — verified by copying the test file alone onto the merge base and running it, not by reasoning. The other three are deliberate regression fences (empty/dot-only still raise; IP literals still return `None`).

The tests use `.de` rather than `.example`, because `.example` is not in the public suffix list and an eTLD+1 assertion against it would pass vacuously.

Full suite on the branch: 5906 passed, 41 skipped, 1 xfailed. ruff, ruff format and mypy clean.

## Not done here

#988 mentions `adagents._idna_ascii_host` as a candidate for deletion. Left alone deliberately — its claimed divergences turned out to be unreachable (whitespace is stripped upstream, `_check_safe_host` runs first, and `idna.encode` already defaults `transitional=False`), so removing it is a cleanup with no behaviour change and belongs in its own PR.

### DIFF
--- a/src/adcp/signing/etld.py
+++ b/src/adcp/signing/etld.py
@@ -20,9 +20,19 @@ snapshot. Bumping the floor on ``tldextract`` is how we refresh.
 
 **Failure-closed convention.** Inputs whose eTLD+1 cannot be derived (raw
 IP addresses, single-label hosts like ``localhost``, hosts that are
-themselves public suffixes) yield ``None`` from :func:`registrable_domain`
-and ``False`` from :func:`same_registrable_domain`. Callers must treat
-None / False as a binding failure, not a soft skip.
+themselves public suffixes, hosts that are not IDNA-encodable — underscore
+labels, labels over 63 bytes, leading/trailing-hyphen labels) yield ``None``
+from :func:`registrable_domain` and ``False`` from
+:func:`same_registrable_domain`. Callers must treat None / False as a
+binding failure, not a soft skip.
+
+**Hosts are compared in canonical A-label form.** ``tldextract`` is
+IDNA-agnostic: hand it a U-label and it hands back a U-label. Comparing
+unencoded hosts made ``straße.de`` and ``xn--strae-oqa.de`` — one host,
+two spellings — never compare equal, refusing a legitimate agent whose
+brand.json and agent URL disagreed on spelling. :func:`host_from`
+therefore delegates to :func:`adcp.signing._idna_canonicalize.canonicalize_host`,
+the same normalizer the JWKS, revocation, and key-origin checks use.
 """
 
 from __future__ import annotations
@@ -30,7 +40,10 @@ from __future__ import annotations
 from functools import lru_cache
 from urllib.parse import urlsplit
 
+import idna
 import tldextract
+
+from ._idna_canonicalize import canonicalize_host
 
 
 @lru_cache(maxsize=1)
@@ -58,15 +71,27 @@ def _extractor() -> tldextract.TLDExtract:
 
 
 def host_from(value: str) -> str:
-    """Return the hostname portion of a URL, or pass a bare host through.
+    """Return the canonical hostname of a URL, or of a bare host.
 
-    Normalizes case and trims a single trailing dot (the FQDN root
-    separator) so ``Example.COM.`` and ``example.com`` compare equal.
+    Both branches delegate to
+    :func:`adcp.signing._idna_canonicalize.canonicalize_host`, which
+    strips a single trailing FQDN-root dot, ASCII-lowercases,
+    short-circuits IPv4/IPv6 literals (IDNA-2008 rejects purely-numeric
+    labels), and otherwise UTS-46 encodes with
+    ``idna.encode(uts46=True, transitional=False)``. So ``Example.COM.``
+    and ``example.com`` compare equal, and so do ``straße.de`` and
+    ``xn--strae-oqa.de``. Routing *both* branches through it is what
+    makes the URL form and the bare-host form of one host normalize
+    identically — previously only the bare-host branch trimmed the root
+    dot, and it trimmed every trailing dot rather than one.
 
     Raises :class:`ValueError` on input that is a URL with no parseable
-    host (``"http://"``) or empty after normalization. URL inputs MUST
-    use a scheme — a bare ``//example.com`` is treated as a bare host,
-    which is by design: bare-host inputs to this helper come from
+    host (``"http://"``), empty after normalization, or not encodable as
+    a hostname. The last case surfaces as ``idna.IDNAError``, which is a
+    :class:`ValueError` subclass (via ``UnicodeError``), so the raise
+    contract is unchanged for callers catching ``ValueError``. URL inputs
+    MUST use a scheme — a bare ``//example.com`` is treated as a bare
+    host, which is by design: bare-host inputs to this helper come from
     ``brand_url`` fields whose schema already constrains them, so a
     bare-host input is never an attacker-controlled URL.
     """
@@ -75,11 +100,13 @@ def host_from(value: str) -> str:
         host = parts.hostname
         if not host:
             raise ValueError(f"URL has no host: {value!r}")
-        return host.lower()
-    stripped = value.strip().rstrip(".").lower()
-    if not stripped:
-        raise ValueError("host is empty")
-    return stripped
+    else:
+        host = value.strip()
+        # Preserve the pre-existing message for ``""`` / ``"."`` / ``".."``;
+        # without this guard ``idna`` would raise "Empty domain" instead.
+        if not host.strip("."):
+            raise ValueError("host is empty")
+    return canonicalize_host(host)
 
 
 def registrable_domain(host_or_url: str) -> str | None:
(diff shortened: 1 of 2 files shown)


bokelley (approved):
Reviewed against current main. The shared IDNA canonicalizer makes U-label/A-label eTLD+1 comparisons converge while preserving the fail-closed behavior for invalid hosts. Focused integration coverage passed.

PR #996, opened 2026-07-29
fix(webhooks): bracket and structurally validate DockerLocalhostRewrite rewrite_to

### LINKED ISSUE #991: DockerLocalhostRewrite.rewrite_to is interpolated into the netloc unvalidated; a bare IPv6 value breaks webhook signing
## Summary

`DockerLocalhostRewrite.rewrite_to` is interpolated straight into a netloc with no validation. An IPv6 value produces an unbracketed authority, which is the one shape RFC 3986 makes ambiguous with a port. This is the only place a non-signing layer synthesizes the netloc that later becomes the signed `@authority`.

## Cause

`src/adcp/webhook_transport_hooks.py:97-100`:

```python
netloc = self.rewrite_to
if parsed.port is not None:
    netloc = f"{self.rewrite_to}:{parsed.port}"
return urlunsplit((parsed.scheme, netloc, parsed.path, parsed.query, parsed.fragment))
```

`rewrite_to` is a plain `str` dataclass field (`:82`) with no constraint beyond its default. The docstring suggests `172.17.0.1` for Linux hosts, so operators are already expected to put an IP literal here; nothing tells them a v6 literal needs brackets, and nothing rejects it if they omit them.

## Reproduce

```python
from adcp.webhook_transport_hooks import DockerLocalhostRewrite
from adcp.signing.canonical import canonicalize_authority

h = DockerLocalhostRewrite(rewrite_to="::1")
h.rewrite_url("https://localhost:9000/hook")
# 'https://::1:9000/hook'

canonicalize_authority("https://::1:9000/hook")
```

| `rewrite_to` | rewritten URL | `@authority` |
|---|---|---|
| `host.docker.internal` | `https://host.docker.internal:9000/hook` | `host.docker.internal:9000` |
| `172.17.0.1` | `https://172.17.0.1:9000/hook` | `172.17.0.1:9000` |
| `::1` | `https://::1:9000/hook` | `TargetUriMalformedError` |
| `2001:db8::1` | `https://2001:db8::1:9000/hook` | `TargetUriMalformedError` |
| `[::1]` | `https://[::1]:9000/hook` | `[::1]:9000` |

Bracketing it by hand works. Nothing tells the operator to.

## Why it matters

The hook runs before signing — `webhook_sender.py:974` calls `apply_hooks`, and the signer sees the rewritten URL — so the ordering is correct and the signature does cover the URL actually dialled. The problem is the value the hook hands forward.

Before #985, `_canon_authority` split the unbracketed netloc on its last colon and reassembled it, so `::1:9000` round-tripped and *looked* fine while being an ambiguous authority. After #985 the same input is a hard `TargetUriMalformedError` at sign time. Failing loudly is the better outcome, but it converts a silent misconfiguration into a delivery failure at first send, with the error raised in `canonical.py` rather than anywhere near the hook that caused it.

Scope is genuinely small: this is an opt-in e2e-testing hook that already refuses to construct unless the sender was built with `allow_private_destinations=True` (`:102-118`), and the failure is at the adopter's own configured destination. It is a config-validation gap, not a protocol defect.

## Suggested direction

`validate_for_sender` (`:102`) is already the construction-time validation seam, and it already raises `ValueError` for the other misconfiguration this hook has. Validate `rewrite_to` there: reject anything that is not a valid hostname or a bracketed/bare IP literal, and bracket a bare IPv6 literal rather than rejecting it — the operator's intent is unambiguous, and the docstring actively encourages IP-literal values.

Doing it at construction keeps the error at wiring time, which is the stated design of this hook's existing check, rather than at the first delivery attempt.

## Related

- #985 — notes this in its own PR body as needing a follow-up ticket; this is that ticket. #985 changes the symptom (silent mis-signing → hard rejection) but does not validate the hook input.

KonstantinMirin:
Fixes #991.

## What was wrong

`DockerLocalhostRewrite.rewrite_to` is interpolated straight into a netloc with no validation. A bare IPv6 value produces an unbracketed authority — the one shape RFC 3986 makes ambiguous with a port:

```
DockerLocalhostRewrite(rewrite_to="::1").rewrite_url("https://localhost:9000/hook")
  before:  https://::1:9000/hook     -> _canon_authority mis-splits at the last colon
  after:   https://[::1]:9000/hook   -> canonicalize_authority() == "[::1]:9000"
```

This is the only place a non-signing layer synthesizes the netloc that later becomes the signed `@authority`, so a mis-split here is a wrong signature rather than a failed request.

## The validation is structural, not syntactic — and that distinction is the point

It rejects only what can move the boundary between authority, userinfo, port, path, query or fragment: path injection, userinfo injection, embedded ports, raw whitespace, control characters, and RFC 6874 zone IDs.

It deliberately does **not** enforce hostname syntax, and I want to flag that as a decision rather than an omission.

The obvious implementation is to route `rewrite_to` through `_idna_canonicalize.canonicalize_host` — the SDK is IDNA-strict nearly everywhere else. That rejects `my_service`, `host_gateway`, `_dns-sd._udp.local`: **legal Docker Compose service names**, resolvable by Docker's embedded DNS, illegal under RFC 952/1123 and IDNA.

`DockerLocalhostRewrite` exists specifically to serve Docker deployments. Rejecting Docker-legal names at *construction* would turn a working deployment into a startup crash on upgrade — and it would do so in the operator-supplied-client path (`webhook_sender.py:1013-1023`), which skips the pinned transport entirely, so those configurations genuinely work today end-to-end. Verified: `canonicalize_authority("https://my_service:9000/hook")` returns `my_service:9000` cleanly on `main`.

Whether a name resolves is the resolver's business. Whether it restructures the signed URL is ours. Non-ASCII names are still IDNA-encoded to A-labels, since those cannot go on the wire as-is.

Measured behaviour:

```
accepted:  my_service, host_gateway, docker_host.local, _dns-sd._udp.local
rejected:  "", attacker.com/path, user@evil.example,
           host.docker.internal:1234, "ho st", fe80::1%eth0
folded:    ::1 -> [::1]   2001:DB8::1 -> [2001:db8::1]
           ::ffff:127.0.0.1 -> [::ffff:7f00:1]   [::1] -> [::1]   172.17.0.1 unchanged
```

## Placement

`validate_for_sender` rather than `__post_init__`, so it also covers direct `apply_hooks` use and hooks not attached to a sender — strictly broader than the issue's suggested direction.

## Adopter-visible change

Values that cannot be represented in a netloc now raise at construction instead of producing a malformed authority at first delivery. Everything the guard rejects was already broken — either rejected downstream by the pinned transport, or silently mis-signed. Nothing that works today stops working; that is the whole reason the guard is structural rather than IDNA-based.

## Test plan

`make ci-local`: **5911 passed, 0 failed, 41 skipped, 1 xfailed.**

25 tests in `tests/conformance/signing/test_webhook_transport_hooks.py`. **9 of them fail** against the pre-fix module — verified by checking `main`'s version of the file over the branch's in a throwaway worktree and re-running, not by reasoning.

One test exists purely to stop a future "tidy-up" from reintroducing the problem: `test_rewrite_to_accepts_docker_legal_service_names` pins the underscore names, so swapping the structural check for an IDNA one fails loudly instead of quietly breaking Docker users.

### DIFF
--- a/src/adcp/webhook_transport_hooks.py
+++ b/src/adcp/webhook_transport_hooks.py
@@ -32,9 +32,17 @@ to the part the use case actually needs.
 
 from __future__ import annotations
 
+import ipaddress
 from dataclasses import dataclass
 from typing import Protocol
 from urllib.parse import urlsplit, urlunsplit
+
+#: Characters that, interpolated into a netloc, move the boundary between
+#: authority / userinfo / port / path / query / fragment -- i.e. rewrite the
+#: signed URL into a different URL. `%` covers RFC 6874 IPv6 zone IDs.
+#: `:` is deliberately absent: IPv6 literals are made of them, so a stray port
+#: is caught after the IP-literal parse instead.
+_STRUCTURAL_CHARS = frozenset('/@?#\\[]%"<>^`{|}')
 
 
 class TransportHook(Protocol):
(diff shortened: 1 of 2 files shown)


bokelley (changes requested):
The construction guard still accepts inputs that normalize to an empty host. Reproduced on current main + this PR:\n\n- DockerLocalhostRewrite(rewrite_to="[]").rewrite_to == ""\n- rewrite_url("https://localhost:9000/hook") returns "https://:9000/hook"\n- DockerLocalhostRewrite(rewrite_to=".").rewrite_to == ""\n\nThe first passes because unwrapping happens after the initial non-empty check; the second passes because the ASCII branch strips all trailing dots without rechecking. Please reject an empty inner value and an empty post-normalization value, with regression coverage for both. The focused suite otherwise passes (29 tests).

KonstantinMirin:
Fixed in `c81fee72`. All three of your repros reject now, and chasing them turned up a fourth shape.

Each reached the empty host past a different check:

- `"[]"` is non-empty until the brackets come off.
- `"."` is non-empty until the trailing root dot comes off.
- `".."` survived my first fix as `"."`, because the strip used `rstrip(".")` and ate *every* trailing dot rather than the one root dot `canonicalize_host` removes.

So the rule is now stated as "no empty label" rather than "not empty", which also covers an interior empty label (`"a..b"`). Regression test over all of them.

You framed it as an empty host; the sharper version is that the assembled URL was `https://:9000/hook` — an authority with a port and no host, which is precisely the shape `@target-uri` canonicalization rejects, produced by the hook whose job is keeping the authority well-formed.

`my_service`, `[::1]`, `172.17.0.1` and `bücher.example` still work. `make ci-local`: 5917 passed, 0 failed.

bokelley (approved):
Verified the follow-up rejects bracket-empty, root-dot-empty, and interior-empty-label rewrite targets while retaining the intended Docker-compatible names. My previous request is fully addressed.

PR #997, opened 2026-07-29
fix(server): unify divergent host normalizers behind one helper

### LINKED ISSUE #990: Three divergent Host-header normalizers: all mangle IPv6 authorities, one leaks userinfo into the tenant key
## Summary

Three separate implementations of "normalize a `Host` header for tenant lookup" exist in the tree. They disagree with each other, all three mangle IPv6 authorities, and one leaks userinfo into the lookup key. One of the three is in `examples/`, which is where adopters copy from.

## The three copies

| | location | body |
|---|---|---|
| A | `src/adcp/server/tenant_router.py:443-454` (`_normalize_host`) | `host.strip().lower()`, then `split(":", 1)[0]` |
| B | `src/adcp/server/tenant_registry.py:227-246` (`_normalize_host`) | `urlparse(raw).netloc or raw`, then `rsplit(":", 1)[0]`, then `.lower()` |
| C | `examples/v3_reference_seller/src/tenant_router.py:58` (inlined in `resolve()`) | `host.strip().lower().split(":", 1)[0]` |

## Reproduce

```python
from adcp.server.tenant_router import _normalize_host as A
from adcp.server.tenant_registry import TenantRegistry
B = TenantRegistry._normalize_host
def C(h): return h.strip().lower().split(":", 1)[0]
```

| input | A (router) | B (registry) | C (example) |
|---|---|---|---|
| `[::1]:8080` | `'['` | `'[::1]'` | `'['` |
| `[::1]` | `'['` | `'[:'` | `'['` |
| `[2001:db8::1]:443` | `'[2001'` | `'[2001:db8::1]'` | `'[2001'` |
| `ACME.example.com:443` | `'acme.example.com'` | `'acme.example.com'` | `'acme.example.com'` |
| `https://user:pw@acme.example.com` | `'https'` | `'user'` | `'https'` |
| `acme.example.com.` | `'acme.example.com.'` | `'acme.example.com.'` | `'acme.example.com.'` |

Three defects visible in that table:

1. **IPv6 authorities are destroyed.** A and C split on the *first* colon, so any bracketed IPv6 `Host` collapses to `'['` or `'[2001'`. B splits on the last colon and so keeps the brackets it should have stripped (`'[::1]'`) — and for a portless `[::1]` produces `'[:'`.
2. **B accepts userinfo into the key.** It uses `.netloc`, not `.hostname`, so `https://user:pw@acme.example.com` normalizes to `'user'`. B's docstring says it accepts full URLs, so this is a reachable input for `resolve_by_host` / `resolve` callers who pass an `agent_url`.
3. **No trailing-dot handling anywhere.** `acme.example.com.` and `acme.example.com` are different tenants in all three.

Row 5 for A and C (`'https'`) is an artifact of feeding them a URL, which their contract does not promise to accept — not a defect on its own, but it shows how far apart the three contracts have drifted while all three are named the same thing.

## Impact

Fail-closed in all three cases: a mangled key matches no tenant row, and the request 404s. There is no cross-tenant resolution here — no realistic tenant is registered under `'['` or `'user'`. So this is a correctness and consistency defect, not an isolation break.

What it costs in practice:

- A deployment reached over an IPv6 literal (health checkers, direct-IP probes, dual-stack ingress that forwards the literal) 404s on every request with no diagnostic pointing at host parsing.
- An adopter who registers a tenant by `agent_url` through `TenantRegistry` and happens to have userinfo in that URL registers it under a garbage key.
- Copy C is in the reference seller example, so the bug propagates into adopter code by design.

## Related hazard, worth deciding before fixing

The obvious fix — route all three through `adcp.signing._idna_canonicalize.canonicalize_host`, which strips brackets, lowercases, strips one trailing dot and A-label-encodes — is **not** a drop-in here. The `Host` header is attacker-controlled, and `canonicalize_host` raises `idna.IDNAError` on underscore hosts and over-long labels. An unwrapped call turns today's clean 404 into an unhandled exception on the ASGI path, i.e. trades a correctness bug for a reachable 500. Whatever the fix is, it needs an explicit `except` that degrades to "no tenant" rather than propagating.

`urlsplit(f"//{host}").hostname` is the cheaper option if the IDNA half is not wanted: it de-brackets IPv6, lowercases, strips the port correctly, and drops userinfo, without introducing a raising path.

## Suggested direction

One shared helper, used by all three sites, with an explicit non-raising contract (returns the best-effort key, never raises). Update the example in the same change so it stops being a divergent copy — otherwise the next adopter reintroduces C.

KonstantinMirin:
Fixes #990.

## What was wrong

Three separate implementations of "normalize a `Host` header for tenant lookup": the subdomain router, the tenant registry, and the v3 reference seller example. They disagreed with each other, all three mangled IPv6 authorities, and one leaked userinfo into the lookup key.

**The IPv6 case is worse than the issue implies.** Splitting on the first colon reduces every bracketed literal to the key `"["`, so two distinct IPv6 tenants collapse onto the same entry. That is cross-tenant mis-resolution, not a fail-closed 404.

```
[::1]:8080            -> ::1
[2001:DB8::0:1]:443   -> 2001:db8::1
user:pw@acme.example.com -> acme.example.com
acme.example.com.     -> acme.example.com
```

## How

One `normalize_host_key`, delegating to `canonicalize_host` — the helper four signing modules already use. Net implementations go from three divergent ones down to one; **no fourth normalizer** in a tree that already had too many. The `examples/` copy is included because it is what adopters clone.

**The helper never raises.** `Host` is attacker-controlled, so a canonicalizer that raised would turn a clean 404 into a 500 — which is exactly why these rows were DEFER rather than MIGRATE in the review that found them. It falls back to the lower-cased input on anything IDNA cannot process. 26 hostile shapes are pinned by test (`''`, `'.'`, `'[::1'`, `'@'`, `'%00.example.com'`, `'http://'`, 100-char labels, underscores), and every output is idempotent under a second pass — load-bearing, because InMemory registration keys are normalized at registration and again at lookup.

## Adopter-visible changes — none isolation-breaking, all worth a release note

**1. Host values that previously 404'd now resolve to the bare host's tenant.**

```
user@acme.example.com   -> acme.example.com
acme.example.com/x      -> acme.example.com
acme.example.com:abc    -> acme.example.com
```

Same tenant in every case, so there is no cross-tenant leak — but it is a Host-header parsing widening, and it matters to anyone whose upstream cache or WAF keys on the raw Host.

**2. UTS-46 folding means compatibility forms now match.** `ＡＣＭＥ.example.com` reaches the tenant registered as `acme.example.com`. Correct IDNA behaviour and consistent with DNS, but if two registration keys fold together `InMemorySubdomainTenantRouter.__init__` silently keeps the last one with no error. That was already true for case variants; the folding enlarges the silently-colliding class.

**3. `import adcp.server` now pulls `adcp.signing`** (~0.13s, 30 modules) because of the module-level `canonicalize_host` import. `idna` is already a core dependency so nothing breaks, and there is precedent (`canonical_formats/references.py` imports the same private module). Per-request normalization cost went 0.1µs → 14µs, which is irrelevant next to a DNS lookup.

## Test plan

`make ci-local`: **5902 passed, 0 failed, 41 skipped, 1 xfailed.**

**5 of the 6 new tests fail** against the old implementations — verified by re-injecting them via a pytest plugin that rebinds `tenant_router._normalize_host` / `normalize_host_key` and `tenant_registry.normalize_host_key`, not by reasoning about the diff. The 6th (`test_normalize_host_key_never_raises_on_hostile_input`) passes either way; it is kept deliberately, because it guards the hazard *this change introduces* — the new `canonicalize_host` call is the thing that could start raising.

## Known wart

`normalize_host_key` strips a trailing dot itself even though `canonicalize_host` already strips one, so `acme.example.com..` double-folds to `acme.example.com`. Harmless for tenant lookup and arguably desirable, but it is a second dot-strip rather than one — noting it rather than leaving it for a reviewer to find.

### DIFF
--- a/examples/v3_reference_seller/src/tenant_router.py
+++ b/examples/v3_reference_seller/src/tenant_router.py
@@ -24,6 +24,9 @@ from sqlalchemy import select
 
 from adcp.server import Tenant
 
+# Not yet re-exported from ``adcp.server``; import from the submodule.
+from adcp.server.tenant_router import normalize_host_key
+
 if TYPE_CHECKING:
     from sqlalchemy.ext.asyncio import async_sessionmaker
 
@@ -49,13 +52,15 @@ class SqlSubdomainTenantRouter:
         self._cache_lock = asyncio.Lock()
 
     async def resolve(self, host: str) -> Tenant | None:
-        # The middleware passes the raw Host header. RFC 7230 makes it
-        # case-insensitive and lets the client include ``:port``; the
-        # Protocol docstring is explicit that implementations strip the
-        # port suffix as needed. Normalize before the cache lookup AND
-        # the DB query so ``acme.localhost:3001`` resolves the same
-        # row as the seeded ``acme.localhost``.
-        host = host.strip().lower().split(":", 1)[0]
+        # The middleware passes the raw Host header. Use the SDK's shared
+        # normalizer rather than hand-rolling a lower-case + port split:
+        # a naive ``split(":", 1)`` mangles bracketed IPv6 authorities,
+        # and rolling your own guarantees it drifts from the key the rest
+        # of the SDK uses. Normalize before the cache lookup AND the DB
+        # query so ``acme.localhost:3001`` resolves the same row as the
+        # seeded ``acme.localhost``. Note the ``tenants.host`` column must
+        # be seeded in this same form (IPv6 de-bracketed, e.g. ``::1``).
+        host = normalize_host_key(host)
         # Bounded FIFO cache — when full, the oldest insertion is
         # evicted regardless of access frequency. Fine for stable
         # tenant sets under ``cache_size``; adopters with churn or
--- a/src/adcp/server/tenant_registry.py
+++ b/src/adcp/server/tenant_registry.py
@@ -35,7 +35,8 @@ import logging
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Literal
-from urllib.parse import urlparse
+
+from adcp.server.tenant_router import normalize_host_key
 
 if TYPE_CHECKING:
     from adcp.decisioning.accounts import AccountStore
@@ -226,10 +227,18 @@ class TenantRegistry:
 
     @staticmethod
     def _normalize_host(raw: str) -> str:
-        """Lower-case and strip any port suffix from a host or URL.
+        """Reduce a host or URL to its tenant-lookup key.
 
         Accepts both full URLs (``https://acme.example.com``) and raw
         Host-header values (``acme.example.com``, ``acme.example.com:443``).
+        Delegates to :func:`~adcp.server.tenant_router.normalize_host_key`
+        so that a tenant registered by ``agent_url`` is reachable by the
+        ``Host`` header the subdomain router resolves — the two used to
+        key the same address differently.
+
+        Beyond lower-casing and port stripping this discards any
+        ``user:pw@`` userinfo, removes IPv6 brackets (``[::1]:8443`` is
+        keyed as ``::1``), and folds a trailing FQDN-root dot.
 
         Note: port stripping is correct for ``Host`` headers where the port
         matches the scheme default. Some load-balancers forward
@@ -237,13 +246,7 @@ class TenantRegistry:
         using that header should strip the port themselves before passing
         the value to :meth:`resolve_by_host` or :meth:`resolve`.
         """
-        if "://" in raw:
-            host = urlparse(raw).netloc or raw
-        else:
-            host = raw
-        if ":" in host:
-            host = host.rsplit(":", 1)[0]
-        return host.lower()
+        return normalize_host_key(raw)
 
     async def _run_validator(self, tenant_id: str) -> bool:
         """Invoke the configured validator; return True when valid."""
--- a/src/adcp/server/tenant_router.py
+++ b/src/adcp/server/tenant_router.py
@@ -501,6 +501,21 @@ def normalize_host_key(value: str) -> str:
         host = None
     if not host:
         host = raw.lower()  # unparseable authority -> best-effort key
+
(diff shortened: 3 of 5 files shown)


bokelley (changes requested):
normalize_host_key does not canonicalize an IPv6 literal after extracting it from a port-bearing authority. Reproduced on current main + this PR:\n\nnormalize_host_key("[2001:DB8::0:1]:443") == "2001:db8::0:1"\nnormalize_host_key("2001:db8::0:1") == "2001:db8::1"\n\nSo the advertised idempotency invariant is false, and a tenant registered under bare [2001:DB8::0:1] is unreachable from the equivalent Host value with :443. Please run parts.hostname through ipaddress.ip_address before the ASCII fast path and add a non-canonical, port-bearing IPv6 regression test. The focused suite otherwise passes (95 tests).

KonstantinMirin:
Fixed in `fe0cceb8`. Your repro was exact.

The bare-literal short-circuit only inspects the raw input. Once a port is attached it is `urlsplit` that removes the brackets, so the address reached the rest of the function uncompressed and was never folded — hence `[2001:DB8::0:1]:443` keying to `2001:db8::0:1` while the bare form keyed to `2001:db8::1`.

Worth being explicit that this makes the docstring's idempotency claim false, not merely incomplete. That matters more than the lookup miss: `InMemorySubdomainTenantRouter` normalizes registration keys at construction and normalizes the Host again at lookup, so a non-idempotent key silently fails to match itself.

Now re-runs the IP-literal test on the extracted host. Two of the four new tests fail without the change (verified by stashing only the source file — stashing everything removes the tests too and proves nothing); the other two are regression guards and I've said so rather than counting them as coverage.

`make ci-local`: 5906 passed, 0 failed.

bokelley (approved):
Verified the extracted host is re-run through IP canonicalization, making port-bearing non-canonical IPv6 values converge with their bare form. My previous request is fully addressed.

PR #998, opened 2026-07-29
fix(signing): canonicalize both origins in the brand.json jwks gate

### LINKED ISSUE #989: brand.json resolver: _canonicalize_url drops IPv6 brackets, and the jwks_origin_mismatch gate compares a canonical origin to a raw one
## Summary

Two origin-normalization defects in the `brand.json` resolver, both from hand-rolled netloc string work:

1. `_canonicalize_url` rebuilds the netloc from `parts.hostname`, which strips IPv6 brackets and never re-adds them. A bracketed IPv6 `brand.json` URL is turned into a malformed URL, and the failure surfaces as an untyped `SSRFValidationError` escaping the resolver instead of a `BrandJsonResolverError`.
2. The `jwks_origin_mismatch` gate compares a *canonicalized* brand origin against a *raw* agent origin, so cosmetic spelling differences in `agent.url` produce a spurious mismatch with an error message that blames a cross-origin trust pivot.

Both fail closed. Neither is a bypass — see "Direction" below, which matters because the second one sits on a security gate and reads worse than it is.

## 1. `_canonicalize_url` de-brackets IPv6 authorities

`src/adcp/signing/brand_jwks.py:714-721`:

```python
host = parts.hostname or ""            # :714 — strips [ ] from IPv6
...
netloc = host if port is None else f"{host}:{port}"   # :720 — never re-adds them
return urlunsplit((scheme, netloc, parts.path, parts.query, ""))
```

```python
>>> from adcp.signing.brand_jwks import _canonicalize_url
>>> _canonicalize_url("https://[::1]/x", allow_private=True)
'https://::1/x'
```

That string is then both the redirect-loop `seen` key (`:551-553`) and the URL handed to the transport builder at `:569` and to `client.get` at `:580`. The builder rejects it:

```python
>>> from adcp.signing.ip_pinned_transport import build_async_ip_pinned_transport
>>> build_async_ip_pinned_transport("https://::1/x", allow_private=True)
adcp.signing.jwks.SSRFValidationError: URI has no host: 'https://::1/x'
```

`build_async_ip_pinned_transport` at `:569` is **outside** the `try` opened at `:577`, and the `except SSRFValidationError` at `:581` only wraps `client.get`. So the exception escapes `_fetch_brand_json` untyped — callers that handle `BrandJsonResolverError` see an `SSRFValidationError` instead. (httpx would have refused the same URL with `httpx.InvalidURL`, which is not an `httpx.HTTPError` and so is not caught by `:585` either.)

Marginal in frequency — `brand.json` URLs are normally DNS names — but the diagnostic is wrong and the exception type is not part of the resolver's contract.

**Same function, second gap.** Its docstring (`:685-701`) states it exists to close "trivial aliasing attacks (case-mismatched host, default-port elision)" against the `seen` set, and that a Python-only deployment would otherwise "fail open where JS fails closed". It does not strip the trailing FQDN-root dot and does not IDNA-encode, so it does not close those two aliases:

```python
>>> _canonicalize_url("https://Brand.Example./x", allow_private=True)
'https://brand.example./x'                 # distinct `seen` entry from .../x
>>> _canonicalize_url("https://bücher.example/x", allow_private=True)
'https://bücher.example/x'                 # distinct from the xn-- spelling
```

`max_redirects` still bounds the chain, so this is not an unbounded loop — it means the depth cap, not the loop detector, is what stops it. Worth closing so the function meets its own stated contract, not because it is currently exploitable.

## 2. The `jwks_origin_mismatch` gate compares canonical to raw

`src/adcp/signing/brand_jwks.py:895-898`:

```python
brand_parts = urlsplit(final_brand_url)
agent_origin = f"{agent_parts.scheme}://{agent_parts.netloc}"
brand_origin = f"{brand_parts.scheme}://{brand_parts.netloc}"
if agent_origin != brand_origin:
    raise BrandJsonResolverError("jwks_origin_mismatch", ...)
```

`final_brand_url` is the output of `_canonicalize_url` (`:548`, `:641`, `:659` → `:673`): host lowercased, default port stripped, fragment stripped, userinfo rejected. `agent_url` is the raw `agents[].url` string from the document. So this is a byte comparison between a canonical value and an uncanonicalized one, and it uses raw `.netloc` — the one accessor in the module that keeps userinfo.

```python
>>> from adcp.signing.brand_jwks import _default_jwks_uri
>>> BRAND = "https://brand.example/.well-known/brand.json"   # already canonicalized

>>> _default_jwks_uri("https://brand.example/agent", BRAND)
'https://brand.example/.well-known/jwks.json'

>>> _default_jwks_uri("https://Brand.Example/agent", BRAND)
BrandJsonResolverError: jwks_origin_mismatch
>>> _default_jwks_uri("https://brand.example:443/agent", BRAND)
BrandJsonResolverError: jwks_origin_mismatch
>>> _default_jwks_uri("https://brand.example./agent", BRAND)
BrandJsonResolverError: jwks_origin_mismatch
>>> _default_jwks_uri("https://user@brand.example/agent", BRAND)
BrandJsonResolverError: jwks_origin_mismatch
```

The IDN pair diverges the same way: a `brand.json` fetched from `https://bücher.example/...` produces `brand_origin == "https://bücher.example"`, so an `agent.url` of `https://xn--bcher-kva.example/agent` mismatches despite naming the same host.

### Direction: this fails closed, and cannot fail open

Stating this explicitly because the site is a security gate and the shape invites the opposite reading.

- The brand side is always the output of `_canonicalize_url`: lowercase host, no userinfo, no default port, no fragment. The agent side is raw. A normalization divergence can therefore only make a raw string *unequal* to a canonical one. There is no input that makes two origins denoting **different** hosts compare equal, because equality here is byte equality against a value that is already in canonical form.
- The return value at `:907` is built from `agent_origin`, but only on the equality branch — where `agent_origin` is byte-identical to `brand_origin`. So the derived `jwks_uri` is always rooted at the canonical brand origin. No trust pivot is reachable through this path.

The honest severity is therefore **interop / diagnostics, low**: a publisher who writes `agent.url: "https://Brand.Example/agent"` or includes an explicit `:443` in a `brand.json` served from `https://brand.example/` loses the default-`jwks_uri` path, and the error text tells them their agent is on a different origin from their brand.json — which it is not. The documented escape hatch (declare an explicit `jwks_uri`) works, so this is recoverable, just confusing.

## Suggested direction

Both findings have the same fix shape: derive the origin through the package's designated host canonicalizer rather than string-concatenating a netloc.

- `_canonicalize_url`: re-bracket the host when it is an IPv6 literal before reassembling the netloc, and route the host through `_idna_canonicalize.canonicalize_host` (which strips one trailing dot, lowercases, short-circuits IP literals, and A-label-encodes). Note `canonicalize_host` returns IPv6 literals unbracketed, so bracket re-addition is the caller's job either way.
- `_default_jwks_uri`: build both origins the same way — `scheme.lower()` + `canonicalize_host(parts.hostname)` + non-default port — rather than comparing raw `.netloc` on one side and a canonical string on the other. Using `.hostname` rather than `.netloc` also drops the userinfo that only the agent side can currently carry.
- Also move `build_async_ip_pinned_transport` at `:569` inside the `try` at `:577`, or wrap it, so an SSRF refusal on the transport-construction path surfaces as `BrandJsonResolverError("fetch_failed", ...)` like the one on the request path already does.

`canonicalize_host` raises on underscore hosts and over-long labels where these functions return them today; both sites are already fail-closed (they raise `BrandJsonResolverError` on bad input), so mapping the raise onto `invalid_url` is consistent with what is there.

KonstantinMirin:
Fixes #989 — both halves.

## Two defects, and they are the same defect from two sides

**1. `_canonicalize_url` dropped IPv6 brackets.** It rebuilt the authority from `urlsplit(...).hostname`, which returns IPv6 literals **de-bracketed**:

```
https://[::1]/x            ->  https://::1/x              (no parseable host)
https://[2001:db8::1]:8443/x ->  https://2001:db8::1:8443/x (port no longer separable)
```

Both are fed straight to `build_async_ip_pinned_transport` on every redirect hop, so the first was refused outright and the second re-parsed wrongly.

**2. `_default_jwks_uri` compared a canonical origin against a raw one.** `final_brand_url` has been through the canonicalizer; `agent.url` is raw as published. So a publisher spelling the *same* origin identically on both sides was told their agent lives somewhere else:

| agent.url / brand.json (same origin, same spelling) | before | after |
|---|---|---|
| `https://Brand.Example/…` | `jwks_origin_mismatch` | ✅ defaults |
| `https://brand.example:443/…` | `jwks_origin_mismatch` | ✅ defaults |
| `https://brand.example./…` | `jwks_origin_mismatch` | ✅ defaults |
| `https://user@brand.example/…` | `jwks_origin_mismatch` | ✅ defaults |
| `https://bücher.example/…` | `jwks_origin_mismatch` | ✅ defaults |
| `https://evil.example/…` vs `brand.example` | `jwks_origin_mismatch` | **still rejected** |

That is exactly the confusing diagnostic the gate exists to avoid — produced by the gate itself.

## Why they had to be fixed together

**Fixing only #1 makes #2 worse.** A stronger canonicalizer on one side of an asymmetric comparison *widens* the gap. An earlier draft of this change did only the bracket fix, and it broke three publisher configurations that work on `main` today — a U-label on both sides, a trailing root dot on both sides, and an underscore host. Anyone reviewing the bracket fix in isolation should know it is not safely separable.

Both sides now run through one `_canonical_origin`, sharing host handling with `_canonicalize_url`: same `canonicalize_host`, same re-bracketing, same default-port elision.

It is built from **`.hostname`, not `.netloc`**. `.netloc` is the one accessor that retains userinfo, so `https://user@brand.example` compared unequal to `https://brand.example` — a trust decision turning on a credential that is not part of an origin at all.

## The security fence is unmoved

An attacker-controlled brand.json naming `agent.url` on another origin is still rejected with `jwks_origin_mismatch`, and `test_cross_origin_agent_is_still_rejected` pins it. **Canonicalizing closes spelling differences, not origin differences** — worth stating explicitly, because "we made an origin comparison more permissive" is the shape of a real vulnerability and this one is not.

Also in scope, from the issue's suggested direction: `build_async_ip_pinned_transport` moves inside the fetch `try` block, so a transport-side SSRF refusal surfaces as `fetch_failed` instead of escaping the resolver as a raw `SSRFValidationError`.

## On the issue's severity framing

#989 describes the origin comparison as a diagnostics/interop problem, not a trust-pivot, and that is correct — I re-verified it. The comparison is byte equality against an already-canonical value, so a normalization divergence can only make the raw side *unequal*; no input makes two different origins compare equal, and the defaulted `jwks_uri` is built on the equality branch where both strings are identical. The user-visible cost was publishers losing the default-`jwks_uri` path and being told something false about their own configuration.

## Test plan

`make ci-local`: **5913 passed, 0 failed, 41 skipped, 1 xfailed.**

17 tests in `tests/conformance/signing/test_brand_jwks_canonicalization.py`.

For the origin-gate tests specifically: **4 of the new ones fail** with `src/adcp/signing/brand_jwks.py` reverted to base and the test file kept — verified by stashing only the source file, since stashing everything removes the tests too and proves nothing. The remaining rows (`u-label`, `ipv6`, and the cross-origin fence) pass either way and are regression guards rather than new coverage; calling that out rather than claiming nine red tests.

### DIFF
--- a/src/adcp/signing/brand_jwks.py
+++ b/src/adcp/signing/brand_jwks.py
@@ -45,7 +45,9 @@ from typing import Any, ClassVar, Literal
 from urllib.parse import urlsplit, urlunsplit
 
 import httpx
+import idna
 
+from adcp.signing._idna_canonicalize import canonicalize_host
 from adcp.signing.jwks import (
     AsyncCachingJwksResolver,
     AsyncJwksFetcher,
@@ -566,7 +568,17 @@ async def _fetch_brand_json(
         if client_factory is not None:
             client_cm = client_factory(url)
         else:
-            transport = build_async_ip_pinned_transport(url, allow_private=allow_private)
+            # The transport builder resolves + validates the host up
+            # front, so it — not the later request — is where an SSRF
+            # refusal surfaces. It must sit inside the same handler as
+            # the request or the refusal escapes the resolver's
+            # documented error contract as a raw SSRFValidationError.
+            try:
+                transport = build_async_ip_pinned_transport(url, allow_private=allow_private)
+            except SSRFValidationError as exc:
+                raise BrandJsonResolverError(
+                    "fetch_failed", f"brand.json URL failed SSRF check: {exc}"
+                ) from exc
             client_cm = httpx.AsyncClient(
                 transport=transport,
                 timeout=timeout_seconds,
@@ -691,6 +703,19 @@ def _canonicalize_url(raw: str, *, allow_private: bool) -> str:
 
     * Scheme lowercased.
     * Host lowercased (``urlsplit`` does NOT do this — we do it).
+    * Trailing FQDN-root dot stripped (``brand.example.`` and
+      ``brand.example`` are the same host).
+    * Unicode U-labels encoded to A-labels (``bücher.example`` →
+      ``xn--bcher-kva.example``), via the package-wide UTS#46
+      convention in :mod:`adcp.signing._idna_canonicalize`. A host the
+      IDNA encoder refuses (e.g. an underscore label) is rejected with
+      ``invalid_url`` rather than passed through.
+    * IP literals normalized to their canonical form, and IPv6
+      literals re-bracketed. ``urlsplit(...).hostname`` returns IPv6
+      hosts *de-bracketed*, so rebuilding the authority from it
+      without re-adding brackets emits ``https://::1/x`` — a string
+      with no parseable host, and with a non-default port a string
+      whose port can no longer be separated from the address.
     * Default port (443 for https, 80 for http) stripped.
     * Fragments stripped — they aren't sent on the wire and must not
       smuggle loop-detection aliases into ``seen``.
@@ -699,6 +724,10 @@ def _canonicalize_url(raw: str, *, allow_private: bool) -> str:
     ``https://x.example/`` as distinct strings; the JS-side resolver
     canonicalizes both via ``new URL``, so a Python-only deployment
     would fail open where JS fails closed.
+
+    The returned string is required to be a URL the transport layer
+    accepts: it is fed straight to ``build_async_ip_pinned_transport``
+    and ``client.get`` on every redirect hop.
     """
     try:
         parts = urlsplit(raw)
@@ -711,9 +740,21 @@ def _canonicalize_url(raw: str, *, allow_private: bool) -> str:
     scheme = parts.scheme.lower()
     if scheme != "https" and not (allow_private and scheme == "http"):
         raise BrandJsonResolverError("invalid_url", "brand.json URL must use https://")
-    host = parts.hostname or ""
-    if not host:
+    raw_host = parts.hostname or ""
+    if not raw_host:
         raise BrandJsonResolverError("invalid_url", "brand.json URL has no host")
+    try:
+        host = canonicalize_host(raw_host)
+    except (idna.IDNAError, UnicodeError) as exc:
+        raise BrandJsonResolverError(
+            "invalid_url", f"brand.json URL host is not a valid IDNA name: {exc}"
+        ) from exc
+    # ``canonicalize_host`` returns IPv6 literals UNBRACKETED (its step
+    # 3 short-circuits to ``str(ipaddress.ip_address(...))``); putting
+    # the brackets back is the caller's job. A canonicalized DNS name
+    # never contains ':', so this is an unambiguous IPv6 test.
+    if ":" in host:
+        host = f"[{host}]"
     port = parts.port
     if port is not None and port == _DEFAULT_PORTS.get(scheme):
         port = None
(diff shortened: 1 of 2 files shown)


bokelley (approved):
Reviewed against current main. IPv6 authorities remain parseable after canonicalization, both sides of the default-JWKS origin comparison now use the same canonical form, and the cross-origin fence remains intact. Focused integration coverage passed.

PR #1000, opened 2026-07-29
fix(decisioning): supervise timed and canceled work

bokelley:
## Summary

- supervise synchronous work that outlives request cancellation or time budgets
- bound timed synchronous admission, including router-backed platforms
- preserve idempotency and proposal lifecycle ownership until underlying work completes

## Why

Timed-out or cancelled requests could release reservations while worker threads were still mutating state. Under load this allowed duplicate work and unbounded executor queues.

## Validation

- focused decisioning, time-budget, proposal, and router saturation tests
- independent concurrency review against current `origin/main`

## Compatibility

- Existing direct asynchronous platform behavior is unchanged.
- BYO `executor=` wiring now requires an explicit `timed_sync_get_products_limit=` because executor wrappers expose no public worker-count contract; framework-allocated pools retain an automatic half-capacity default.
- INTERNAL_ERROR projections no longer include `details.caused_by.message`; they retain only the exception type and the normative `recovery` value. This is wire-compatible with AdCP 3.1.8 because `schemas/cache/3.1/core/error.json` leaves `details` open (`additionalProperties: true`) and does not define `caused_by`.

### DIFF
--- a/src/adcp/decisioning/dispatch.py
+++ b/src/adcp/decisioning/dispatch.py
@@ -59,6 +59,12 @@ from adcp.decisioning.task_registry import (
     TaskHandoffContext,
     TaskRegistry,
 )
+from adcp.decisioning.time_budget import (
+    RoutedSyncExecution,
+    SyncExecutorAdmission,
+    _bind_routed_sync_execution,
+    submit_supervised,
+)
 from adcp.decisioning.types import (
     AdcpError,
     TaskHandoff,
@@ -84,6 +90,11 @@ if TYPE_CHECKING:
     WebhookDeliveryTarget = WebhookSender | WebhookDeliverySupervisor
 
 logger = logging.getLogger(__name__)
+
+# Strong references for synchronous adopter lifecycles that outlive a
+# cancelled request. A Python thread cannot be cancelled; its completion hooks
+# must still settle durable proposal/idempotency state.
+_SUPERVISED_SYNC_LIFECYCLES: set[asyncio.Task[Any]] = set()
 
 # ---------------------------------------------------------------------------
 # Specialism enum — spec slugs known to the framework
@@ -583,6 +594,11 @@ def _internal_error_message(method_name: str, exc: BaseException) -> str:
     return f"Platform method {method_name!r} raised {cls_name}; see details for cause"
 
 
+def _exception_cause_details(exc: BaseException) -> dict[str, Any]:
+    """Return the shared sanitized exception-type breadcrumb."""
+    return {"caused_by": {"type": type(exc).__name__}}
+
+
 def _internal_error_details(exc: BaseException) -> dict[str, Any]:
     """Build the wire-side ``details`` payload for an INTERNAL_ERROR
     wrap.
@@ -624,11 +640,7 @@ def _internal_error_details(exc: BaseException) -> dict[str, Any]:
     where a structured field list is meaningful, so we don't
     generalize this to other exception types.
     """
-    details: dict[str, Any] = {
-        "caused_by": {
-            "type": type(exc).__name__,
-        }
-    }
+    details = _exception_cause_details(exc)
     # Try to import lazily so a future refactor that splits the
     # validation tooling can't ripple through the dispatch layer.
     try:
@@ -647,8 +659,7 @@ def _internal_error_details(exc: BaseException) -> dict[str, Any]:
             details["validation_errors"] = list(narrow_union_errors(errors_list))
         except Exception:
             # Defensive — never let a narrowing bug 500 the wire.
-            # The caused_by.message already carries the truncated
-            # repr; adopters can still triage via server logs.
+            # The exception type still lets adopters triage via server logs.
             pass
     return details
 
@@ -1313,6 +1324,7 @@ async def _invoke_platform_method(
     webhook_target: WebhookDeliveryTarget | None = None,
     webhook_auto_emit: bool = True,
     pre_handoff_reject: Callable[[], None] | None = None,
+    sync_admission: SyncExecutorAdmission | None = None,
 ) -> Any:
     """Invoke a platform method, projecting hybrid returns.
 
@@ -1383,12 +1395,17 @@ async def _invoke_platform_method(
         off on a ``wholesale`` request is rejected cleanly instead of
         leaking a task the buyer was told was rejected. Runs only on the
         ``TaskHandoff`` arm; sync / workflow-handoff returns ignore it.
+    :param sync_admission: Optional bounded admission controller for a sync
+        method. Its permit remains held until the underlying thread future
+        actually completes, including after caller cancellation.
     """
     # pydantic is a required dep; import here (not at module level) to mirror
     # the lazy-import discipline used throughout this module.
     from pydantic import ValidationError as _ValidationError  # noqa: PLC0415
 
     method = getattr(platform, method_name)
+    sync_lifecycle_continues = False
+    routed_sync_execution: RoutedSyncExecution | None = None
     # Re-validate through the platform method's own annotation when it's a
     # stricter subclass of the shim's already-deserialized type.  Skipped
     # when arg_projector is set — that path replaces positional args entirely.
@@ -1407,31 +1424,50 @@ async def _invoke_platform_method(
 
     try:
         if asyncio.iscoroutinefunction(method):
-            if arg_projector is not None:
-                result = await method(**arg_projector, ctx=ctx)
-            elif extra_kwargs:
-                result = await method(params, ctx, **extra_kwargs)
-            else:
-                result = await method(params, ctx)
+            # Async router delegates may resolve to synchronous tenant
+            # children only after account routing. Propagate the same bounded
+            # admission controller and configured executor through ContextVars
+            # so that path cannot bypass the timed-sync limit.
+            with _bind_routed_sync_execution(sync_admission, executor) as routed_sync_execution:
+                if arg_projector is not None:
+                    result = await method(**arg_projector, ctx=ctx)
+                elif extra_kwargs:
+                    result = await method(params, ctx, **extra_kwargs)
+                else:
+                    result = await method(params, ctx)
         else:
-            ctx_snapshot = contextvars.copy_context()
-            loop = asyncio.get_running_loop()
             if arg_projector is not None:
                 projected_kwargs = {**arg_projector, "ctx": ctx}
-                result = await loop.run_in_executor(
-                    executor,
-                    functools.partial(ctx_snapshot.run, method, **projected_kwargs),
-                )
+                worker_call = functools.partial(method, **projected_kwargs)
             elif extra_kwargs:
-                result = await loop.run_in_executor(
-                    executor,
-                    functools.partial(ctx_snapshot.run, method, params, ctx, **extra_kwargs),
-                )
+                worker_call = functools.partial(method, params, ctx, **extra_kwargs)
             else:
-                result = await loop.run_in_executor(
-                    executor,
-                    functools.partial(ctx_snapshot.run, method, params, ctx),
-                )
+                worker_call = functools.partial(method, params, ctx)
+
+            worker_async_future = await submit_supervised(
+                executor,
+                sync_admission,
+                worker_call,
+            )
+            try:
+                result = await asyncio.shield(worker_async_future)
+            except asyncio.CancelledError:
+                if on_complete is not None or on_failure is not None:
+                    sync_lifecycle_continues = True
+                    _supervise_sync_lifecycle(
+                        worker_async_future,
+                        ctx=ctx,
+                        method_name=method_name,
+                        registry=registry,
+                        executor=executor,
+                        on_complete=on_complete,
+                        on_failure=on_failure,
+                        pre_handoff_reject=pre_handoff_reject,
+                        request_params=params,
+                        webhook_target=webhook_target,
+                        webhook_auto_emit=webhook_auto_emit,
+                    )
+                raise
     except AdcpError as exc:
         # Adopter raised structured error — propagate verbatim. The
         # outer middleware projects to the wire envelope. Fire
@@ -1526,8 +1562,8 @@ async def _invoke_platform_method(
         # The ``details.caused_by`` shape (Emma AudioStack P2) gives
         # adopters a breadcrumb on the wire — without it, "An internal
         # error occurred" is a dead end and adopters have to grep
-        # server logs. We expose only the exception class name + str
-        # (not the traceback) so a misconfigured platform that throws
+        # server logs. We expose only the exception class name (not the
+        # message or traceback) so a misconfigured platform that throws
         # on secret material doesn't leak the secret value through
         # the wire response.
         logger.exception(
@@ -1543,7 +1579,64 @@ async def _invoke_platform_method(
         if on_failure is not None:
             await _safe_on_failure_call(on_failure, wrapped, method_name)
         raise wrapped from exc
+    except BaseException as exc:
+        # ``asyncio.CancelledError`` (and shutdown BaseExceptions) bypass the
+        # wire-error wrapping above, but must still release framework state
+        # reserved before adapter dispatch. Preserve the exact exception.
+        nested_sync_future = (
+            routed_sync_execution.worker if routed_sync_execution is not None else None
+        )
+        if isinstance(nested_sync_future, asyncio.Future) and (
+            on_complete is not None or on_failure is not None
+        ):
+            sync_lifecycle_continues = True
+            _supervise_sync_lifecycle(
+                nested_sync_future,
+                ctx=ctx,
+                method_name=method_name,
+                registry=registry,
+                executor=executor,
+                on_complete=on_complete,
+                on_failure=on_failure,
+                pre_handoff_reject=pre_handoff_reject,
+                request_params=params,
+                webhook_target=webhook_target,
+                webhook_auto_emit=webhook_auto_emit,
+            )
+        if on_failure is not None and not sync_lifecycle_continues:
+            await _safe_on_failure_call(on_failure, exc, method_name)
+        raise
 
+    return await _project_invocation_result(
+        result,
+        ctx=ctx,
+        method_name=method_name,
+        registry=registry,
+        executor=executor,
+        on_complete=on_complete,
+        on_failure=on_failure,
+        pre_handoff_reject=pre_handoff_reject,
+        request_params=params,
+        webhook_target=webhook_target,
+        webhook_auto_emit=webhook_auto_emit,
+    )
+
+
+async def _project_invocation_result(
+    result: Any,
+    *,
+    ctx: RequestContext[Any],
+    method_name: str,
+    registry: TaskRegistry,
+    executor: ThreadPoolExecutor,
+    on_complete: Callable[[Any], Awaitable[None]] | None,
+    on_failure: Callable[[BaseException], Awaitable[None]] | None,
+    pre_handoff_reject: Callable[[], None] | None,
+    request_params: BaseModel,
+    webhook_target: WebhookDeliveryTarget | None,
+    webhook_auto_emit: bool,
+) -> Any:
+    """Project a raw adopter result and settle its framework lifecycle hooks."""
     if is_task_handoff(result):
         # Reject before any side effect (registry row, background task,
         # completion webhook) is created. The wholesale discovery guard
@@ -1560,7 +1653,7 @@ async def _invoke_platform_method(
             executor=executor,
             on_complete=on_complete,
             on_failure=on_failure,
-            request_params=params,
+            request_params=request_params,
             webhook_target=webhook_target,
             webhook_auto_emit=webhook_auto_emit,
         )
@@ -1571,7 +1664,7 @@ async def _invoke_platform_method(
             method_name=method_name,
             registry=registry,
             executor=executor,
-            request_params=params,
+            request_params=request_params,
         )
 
     # Sync return path. Fire on_complete with the typed result before
@@ -1594,6 +1687,99 @@ async def _invoke_platform_method(
     # ``extra='allow'``. Method-gated to avoid walking large product
     # / signal catalogs that can't carry credentials.
     return strip_credentials_from_wire_result(method_name, result)
+
+
+async def _settle_cancelled_sync_lifecycle(
+    worker_future: asyncio.Future[Any],
+    *,
+    ctx: RequestContext[Any],
+    method_name: str,
+    registry: TaskRegistry,
+    executor: ThreadPoolExecutor,
+    on_complete: Callable[[Any], Awaitable[None]] | None,
+    on_failure: Callable[[BaseException], Awaitable[None]] | None,
+    pre_handoff_reject: Callable[[], None] | None,
+    request_params: BaseModel,
+    webhook_target: WebhookDeliveryTarget | None,
+    webhook_auto_emit: bool,
+) -> None:
+    """Settle a sync worker after its request task has been cancelled."""
+    try:
+        result = await asyncio.shield(worker_future)
+    except asyncio.CancelledError:
+        # Cancelling this supervisor must not cancel or roll back the
+        # non-cancellable thread it observes. Its reservation remains held.
+        raise
+    except Exception as exc:
+        if on_failure is not None:
+            await _safe_on_failure_call(on_failure, exc, method_name)
+        return
+    if is_task_handoff(result) or is_workflow_handoff(result):
+        # The cancelled caller never received a task id. Do not promote an
+        # unreachable handoff; returning a handoff has not executed its work.
+        if on_failure is not None:
+            await _safe_on_failure_call(on_failure, asyncio.CancelledError(), method_name)
+        logger.warning(
+            "Discarded %s handoff returned after request cancellation; no task id was issued",
+            method_name,
+        )
+        return
+    try:
+        await _project_invocation_result(
+            result,
+            ctx=ctx,
+            method_name=method_name,
+            registry=registry,
+            executor=executor,
+            on_complete=on_complete,
+            on_failure=on_failure,
+            pre_handoff_reject=pre_handoff_reject,
+            request_params=request_params,
+            webhook_target=webhook_target,
+            webhook_auto_emit=webhook_auto_emit,
+        )
+    except Exception:
+        # Lifecycle hooks already apply their own rollback semantics. There is
+        # no request waiter left to receive this exception, so retain it in
+        # server logs rather than producing an unhandled-task warning.
+        logger.exception(
+            "Cancelled request's synchronous %s lifecycle failed while settling",
+            method_name,
+        )
+
+
+def _supervise_sync_lifecycle(
+    worker_future: asyncio.Future[Any],
+    *,
+    ctx: RequestContext[Any],
+    method_name: str,
+    registry: TaskRegistry,
+    executor: ThreadPoolExecutor,
+    on_complete: Callable[[Any], Awaitable[None]] | None,
+    on_failure: Callable[[BaseException], Awaitable[None]] | None,
+    pre_handoff_reject: Callable[[], None] | None,
+    request_params: BaseModel,
+    webhook_target: WebhookDeliveryTarget | None,
+    webhook_auto_emit: bool,
+) -> None:
+    """Own a cancelled request's worker until its lifecycle settles."""
+    lifecycle = asyncio.create_task(
+        _settle_cancelled_sync_lifecycle(
+            worker_future,
+            ctx=ctx,
+            method_name=method_name,
+            registry=registry,
+            executor=executor,
+            on_complete=on_complete,
+            on_failure=on_failure,
+            pre_handoff_reject=pre_handoff_reject,
+            request_params=request_params,
+            webhook_target=webhook_target,
+            webhook_auto_emit=webhook_auto_emit,
+        )
+    )
+    _SUPERVISED_SYNC_LIFECYCLES.add(lifecycle)
+    lifecycle.add_done_callback(_SUPERVISED_SYNC_LIFECYCLES.discard)
 
 
 async def _safe_on_failure_call(
@@ -1808,6 +1994,11 @@ async def _project_handoff(
             )
             await _fail(wrapped)
             return
+        except BaseException:
+            # Cancellation does not prove adopter work stopped. Leave any
+            # reservation fail-closed for expiry/reconciliation rather than
+            # release it while side effects may still be outstanding.
+            raise
 
         # Framework completion hook (e.g., proposal_store.commit for
         # finalize, mark_proposal_consumed for create_media_buy). Runs
--- a/src/adcp/decisioning/handler.py
+++ b/src/adcp/decisioning/handler.py
@@ -55,6 +55,9 @@ from adcp.decisioning.discovery_guards import (
 )
 from adcp.decisioning.dispatch import (
     _build_request_context,
+    _exception_cause_details,
+    _internal_error_details,
+    _internal_error_message,
     _invoke_platform_method,
 )
 from adcp.decisioning.implementation_config import ProductConfigStore
@@ -79,7 +82,11 @@ from adcp.decisioning.refine import (
     has_refine_support,
     project_refine_response,
 )
-from adcp.decisioning.time_budget import project_incomplete_response, resolve_time_budget
+from adcp.decisioning.time_budget import (
+    SyncExecutorAdmission,
+    project_incomplete_response,
+    resolve_time_budget,
+)
 from adcp.decisioning.types import (
     Account as _DecisioningAccount,
 )
@@ -1296,6 +1303,7 @@ class PlatformHandler(ADCPHandler[ToolContext]):
         property_list_fetcher: PropertyListFetcher | None = None,
         media_buy_store: MediaBuyStore | None = None,
         advertise_all: bool = False,
+        timed_sync_get_products_limit: int | None = None,
     ) -> None:
         super().__init__()
         self._platform = platform
@@ -1321,6 +1329,13 @@ class PlatformHandler(ADCPHandler[ToolContext]):
         self.canonical_format_legacy_resolver = getattr(
             platform, "canonical_format_legacy_resolver", None
         )
+        # Direct PlatformHandler construction has no public way to inspect a
+        # BYO executor's capacity. The adopter-facing composition root passes
+        # an explicit resolved value; direct construction defaults to one.
+        admission_limit = (
+            timed_sync_get_products_limit if timed_sync_get_products_limit is not None else 1
+        )
+        self._timed_sync_get_products_admission = SyncExecutorAdmission(admission_limit)
 
         # Cache whether the platform's create_media_buy accepts 'configs'
         # so we only pay the inspect.signature cost at construction time.
@@ -1653,17 +1668,9 @@ class PlatformHandler(ADCPHandler[ToolContext]):
             )
             raise AdcpError(
                 "INTERNAL_ERROR",
-                message=(
-                    "Unhandled exception in platform.get_adcp_capabilities_for_request: "
-                    f"{type(exc).__name__}: {exc}"
-                ),
+                message=_internal_error_message("get_adcp_capabilities_for_request", exc),
                 recovery="terminal",
-                details={
-                    "caused_by": {
-                        "type": type(exc).__name__,
-                        "message": str(exc),
-                    }
-                },
+                details=_internal_error_details(exc),
             ) from exc
         has_scoped_caps = scoped_caps is not None
         if scoped_caps is not None:
@@ -1999,6 +2006,9 @@ class PlatformHandler(ADCPHandler[ToolContext]):
             registry=self._registry,
             on_complete=_persist_draft_hook,
             pre_handoff_reject=pre_handoff_reject,
+            sync_admission=(
+                self._timed_sync_get_products_admission if deadline is not None else None
+            ),
             **self._handoff_webhook_kwargs(),
         )
         try:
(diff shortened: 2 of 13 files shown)


src/adcp/decisioning/dispatch.py
```diff
@@ -1809,6 +1969,12 @@ async def _run() -> None:
             )
             await _fail(wrapped)
             return
+        except BaseException as exc:
```
KonstantinMirin: This releases the proposal reservation while the adopter's handoff work is still outstanding. For `create_media_buy`, `on_failure` is `_release_reservation_hook` -> `release_proposal_reservation` -> CONSUMING -> COMMITTED. Background handoff tasks are detached `create_task`s (`dispatch.py:2069`) that ordinary loop/SIGTERM shutdown cancels. Measured, cancelling after the handoff fn issued its upstream create:

```
HEAD  state after bg cancel: ProposalState.COMMITTED | registry row: submitted
      RETRY ACCEPTED -> {'media_buy_id': 'mb_duplicate_2'}
BASE  state after bg cancel: ProposalState.CONSUMING
      RETRY REJECTED -> AdcpError[PROPOSAL_NOT_COMMITTED / correctable]
```

One committed proposal, two media buys. The arm also skips `_fail()`, so the registry row stays `submitted` with no terminal webhook — the buyer holds a task that never terminates, which is what provokes the retry.

The arm infers "our asyncio task was cancelled" => "the adopter's work did not happen", the inference the rest of this PR refutes on the request path. The handoff fn is not shielded here, and for a sync fn it is a `run_in_executor` thread that cannot be stopped. Either don't fire `on_failure` here (leave CONSUMING, which fails closed and is what eviction is for), or mirror `_settle_cancelled_sync_lifecycle` and settle from the fn's real terminal outcome. A release must be paired with a terminal registry state either way. Lines 1972-1977 never execute under the suite.

Also: "then remain cancellation to the task scheduler" does not parse.

src/adcp/decisioning/dispatch.py
```diff
@@ -1407,31 +1418,74 @@ async def _invoke_platform_method(
...
+                        # Event loop already closed during process teardown.
+                        pass
+
+                worker_future.add_done_callback(_release_admission)
+
+            worker_async_future = asyncio.wrap_future(worker_future, loop=loop)
+            try:
+                result = await asyncio.shield(worker_async_future)
+            except asyncio.CancelledError:
+                if on_failure is not None:
```
KonstantinMirin: This predicate reads one half of the `(on_complete, on_failure)` lifecycle contract. `get_products` is the only method with a deadline and the only caller passing `sync_admission`, and it wires `on_complete=_persist_draft_hook` with no `on_failure` (`handler.py:1974-1979`), so no timed `get_products` is ever supervised:

```
on_complete only     -> ON_COMPLETE CALLS: []
add on_failure=noop  -> ON_COMPLETE CALLS: [{'products': []}]
```

The else-branch also leaves the shielded future with no owner, so a worker raising after the deadline reaches asyncio's default handler. New at HEAD, absent at `b7ef1dfc`:

```
ERROR:asyncio:Future exception was never retrieved
future: <Future finished exception=RuntimeError('upstream token sk-SECRET-LEAK rejected')>
```

That is the adopter's raw exception text at ERROR, outside the framework's error funnel, while this same PR strips `str(exc)` from `details.caused_by` for exactly that reason.

Gate on `on_complete is not None or on_failure is not None`, and attach a consumer to `worker_async_future` in every branch. Raised in the 2026-07-29 review as a follow-up; still open.

src/adcp/decisioning/dispatch.py
```diff
@@ -1543,15 +1597,77 @@ async def _invoke_platform_method(
         if on_failure is not None:
             await _safe_on_failure_call(on_failure, wrapped, method_name)
         raise wrapped from exc
+    except BaseException as exc:
+        # ``asyncio.CancelledError`` (and shutdown BaseExceptions) bypass the
+        # wire-error wrapping above, but must still release framework state
+        # reserved before adapter dispatch. Preserve the exact exception.
+        nested_sync_future = getattr(exc, "_adcp_sync_worker_future", None)
+        if isinstance(nested_sync_future, asyncio.Future) and on_failure is not None:
```
KonstantinMirin: Same predicate as `dispatch.py:1469`, and the same consequence on the routed path: `platform_router.py:166` sets `_adcp_sync_worker_future` on every cancelled routed sync call, but this is its only consumer, so on the `get_products` path the marker is written and never read.

This block is also byte-identical to `1471-1487` bar the future it settles. Extract one `_supervise_sync_lifecycle(future, ...)` so the predicate lives in a single place — two verbatim copies is what let both share the wrong condition.

src/adcp/decisioning/dispatch.py
```diff
@@ -1407,31 +1418,74 @@ async def _invoke_platform_method(
...
+                    ctx_snapshot.run, method, params, ctx, **extra_kwargs
                 )
             else:
-                result = await loop.run_in_executor(
-                    executor,
-                    functools.partial(ctx_snapshot.run, method, params, ctx),
-                )
+                worker_call = functools.partial(ctx_snapshot.run, method, params, ctx)
+
+            if sync_admission is not None:
```
KonstantinMirin: This acquire/submit/release/wrap/shield sequence is duplicated verbatim in `platform_router.py:141-162` — same steps, same order, differing only in identifiers and indentation, and already diverging on the fallback branch and on cancellation. Every step is load-bearing for permit accounting, so a fix applied to one copy silently misses the other.

`SyncExecutorAdmission` (`time_budget.py:82`) exposes only `acquire`/`release` and leaves the ordering contract — release exactly once, from the concurrent future's callback, marshalled back to the loop — to be re-derived at each call site. Extract `submit_supervised(executor, admission, call) -> asyncio.Future` next to the semaphore. The third framework sync-submit site, `proposal_dispatch.py:244`, is still the pre-PR bare `run_in_executor` and would be its third caller.

src/adcp/decisioning/dispatch.py
```diff
@@ -1543,15 +1597,77 @@ async def _invoke_platform_method(
...
         # completion webhook) is created. The wholesale discovery guard
         # uses this so an adopter handing off on a synchronous-only
         # wholesale request never leaks a task the buyer is told is
         # rejected.
         if pre_handoff_reject is not None:
-            pre_handoff_reject()
+            try:
+                pre_handoff_reject()
+            except BaseException as exc:
+                if on_failure is not None:
```
KonstantinMirin: This branch is dead by construction. `pre_handoff_reject=` is passed by exactly two callers (`handler.py:1976` `get_products`, `handler.py:2676` `get_signals`); `on_failure=` by exactly one (`handler.py:2192` `create_media_buy`). No caller passes both, and coverage confirms line 1669 never executes.

Same root as the supervision gate: the framework's two discovery tools carry a completion hook and no failure hook. Give them one, or hoist supervision off the `on_failure` predicate — then this branch becomes reachable and testable.

src/adcp/decisioning/dispatch.py
```diff
@@ -1610,10 +1771,9 @@ async def _safe_on_failure_call(
     """
     try:
         await on_failure(exc)
-    except Exception:
+    except BaseException:
```
KonstantinMirin: Widening to `BaseException` here swallows a `CancelledError` delivered while `on_failure` is awaited. `on_failure` hooks do durable-store I/O, so they contain await points where cancellation lands; when it does, the hook's `CancelledError` is logged and discarded and the caller re-raises the original exception. Adopter raises `AdcpError`, request task cancelled mid store round-trip:

```
HEAD  RESULT: cancellation SWALLOWED, task raised AdcpError[INTERNAL_ERROR / terminal]
      cancelled() -> False | reservation released -> []
BASE  RESULT: task honoured cancellation
      cancelled() -> True  | reservation released -> []
```

A graceful-shutdown drain doing `task.cancel(); await task` gets a completed task back, and the release is left half-done anyway. Log-and-continue is right for `Exception`; for a `BaseException` that is not an `Exception`, log and `raise`. Line 1774 never executes under the suite.

src/adcp/decisioning/proposal_dispatch.py
```diff
@@ -744,18 +744,15 @@ async def maybe_hydrate_recipes_for_create_media_buy(
...
                 field_path_prefix="packages",
             )
-    except Exception:
-        # Narrow to Exception (not BaseException): CancelledError /
-        # SystemExit / KeyboardInterrupt skip the release path — under
-        # cancellation the next worker will read the stale reservation
-        # and eviction handles it; under shutdown we want fast exit.
-        # release_consumption is idempotent on already-COMMITTED so a
-        # release that races with a concurrent worker is harmless.
+    except BaseException:
```
KonstantinMirin: The comment's stated trigger cannot happen. The `try:` this guards (lines 698-746) contains no `await` — `_derive_packages`, the `params.packages` assignment and `validate_capability_overlap` are all synchronous — so `CancelledError` can never be delivered inside it. The only reachable non-`Exception` cases are a synchronous `SystemExit`/`KeyboardInterrupt` during derivation.

Either narrow this back to `Exception` and drop the comment, or state the reachable trigger and write the test for it. As written the block cannot be graded.

src/adcp/decisioning/proposal_dispatch.py
```diff
@@ -888,7 +885,7 @@ async def release_proposal_reservation(
                 expected_account_id=proposal_record.account_id,
             )
         )
-    except Exception:
+    except BaseException:
```
KonstantinMirin: Same shape as `dispatch.py:1774`: a `CancelledError` raised from `store.release_consumption` is now logged and swallowed where it previously propagated. The caller then continues as if the cancellation never arrived. Log-and-continue is right for `Exception`; re-raise a `BaseException` that is not an `Exception` after the cleanup. Line 888 never executes under the suite — add a store double whose `release_consumption` raises `asyncio.CancelledError` and pin the contract you now depend on.

src/adcp/decisioning/handler.py
```diff
@@ -1298,6 +1305,16 @@ def __init__(
         self._property_list_fetcher = property_list_fetcher
         self._media_buy_store = media_buy_store
         self._advertise_all = advertise_all
+        # ThreadPoolExecutor exposes no public worker-count property. Standard
+        # instances carry ``_max_workers``; executor-compatible wrappers fall
+        # back to the conservative one-call limit.
+        worker_count = int(getattr(executor, "_max_workers", 1))
```
KonstantinMirin: `_max_workers` is a CPython-internal field, and `serve.py:120-124` documents BYO executors as "for operators with audit-instrumented thread pools or **wrappers around stdlib's executor**" — the shape that has no `_max_workers`:

```
wrapper around a 64-worker pool -> admission limit 1
plain 64-worker pool            -> admission limit 32
```

Every deadline-managed sync `get_products` in that deployment then serialises behind one permit held for the full worker duration, with no warning and no public way to read the value. The parameter is annotated `executor: ThreadPoolExecutor` (`handler.py:1278`), so by the declared type the `getattr` default is dead code; for the shape the docs advertise it is a silent cliff. `int()` on an executor whose `_max_workers` is not int-able raises at handler construction.

The composition root already resolved this number (`serve.py:70,268`). Thread the resolved worker count (or the resolved limit) into `PlatformHandler` and keep `timed_sync_get_products_limit` as the explicit override; if a BYO executor cannot supply a count, require the override rather than silently choosing 1.

src/adcp/decisioning/platform_router.py
```diff
@@ -128,6 +131,42 @@
     from adcp.decisioning.proposal_store import ProposalStore
 
 
+async def _run_sync_delegate(method: Any, *args: Any, **kwargs: Any) -> Any:
+    """Run a sync child and expose its live future across request cancellation."""
+    admission, executor = _routed_sync_execution()
+    if admission is None or executor is None:
```
KonstantinMirin: `_bind_routed_sync_execution(sync_admission, executor)` always binds a non-`None` executor (`dispatch.py:1425`), but this condition only honours it when an admission controller is also present — i.e. only for deadline-managed `get_products`. Every other routed sync child (`create_media_buy`, `update_media_buy`, `refine_get_products`, all synthesized delegates) runs on the loop's default executor with the right executor sitting in the ContextVar one line above. With a BYO pool named `FRAMEWORK`, a routed sync child on a no-deadline `get_products` ran on thread `asyncio_0`, which undercuts the D5 BYO-executor contract documented in `serve.py`.

Two orthogonal decisions are collapsed into one condition. Split them: `executor is not None` -> submit to it; `admission is not None` -> additionally gate on a permit. Add a test asserting a routed sync child runs on the handler's executor (`thread_name_prefix`) with and without a time budget.

src/adcp/decisioning/platform_router.py
```diff
@@ -128,6 +131,42 @@
...
+                pass
+
+        concurrent_worker.add_done_callback(_release_admission)
+        worker = asyncio.wrap_future(concurrent_worker, loop=loop)
+    try:
+        return await asyncio.shield(worker)
+    except asyncio.CancelledError as exc:
+        # dispatch._invoke_platform_method consumes this private marker and
+        # settles lifecycle hooks from the worker's actual terminal outcome.
+        setattr(exc, "_adcp_sync_worker_future", worker)
```
KonstantinMirin: This encodes dispatch's settlement protocol as a private attribute on an exception instance, read back at `dispatch.py:1604` as `Any` and narrowed only by `isinstance(..., asyncio.Future)`. `platform_router` is a `DecisioningPlatform` implementation below dispatch, so the dependency is inverted, and there is no type contract — nothing fails at type-check time if the attribute name, producer or consumer drifts.

The channel is also lossy: any frame between the router and dispatch that catches `CancelledError` and re-raises a fresh instance (adopter middleware, a wrapping platform) drops the marker with no log at either end, and the worker is orphaned.

The two modules already share a typed, scoped channel — the `_bind_routed_sync_execution` / `_routed_sync_execution` pair. Put the live worker future on that scope object, or on a `SyncWorkerCancelled(CancelledError)` subclass with a typed `worker: asyncio.Future[Any]`. Either way, log when dispatch unwinds a cancelled routed sync call with no marker.

src/adcp/decisioning/serve.py
```diff
@@ -371,6 +378,7 @@ def create_adcp_server_from_platform(
         property_list_fetcher=property_list_fetcher,
         media_buy_store=media_buy_store,
         advertise_all=advertise_all,
+        timed_sync_get_products_limit=timed_sync_get_products_limit,
```
KonstantinMirin: Deleting this line and its twin at `serve.py:588` leaves the whole suite green — the new public parameter becomes a documented no-op and nothing notices. `timed_sync_get_products_limit` appears in tests only as a direct `PlatformHandler(...)` kwarg (`tests/test_time_budget.py:312,393,447`); neither adopter-facing entry point is exercised.

Add `create_adcp_server_from_platform(platform, timed_sync_get_products_limit=1)` and assert the resulting handler's admission limit, plus a handler built on `ThreadPoolExecutor(max_workers=4)` with no override that admits 2 and blocks the 3rd.

src/adcp/decisioning/time_budget.py
```diff
@@ -61,15 +63,74 @@
...
+
+    A permit represents a worker submission, not a waiting HTTP request. It
+    is released only by the underlying ``concurrent.futures.Future`` done
+    callback, because cancelling its asyncio wrapper cannot stop a running
+    Python thread.
+    """
+
+    def __init__(self, limit: int) -> None:
+        if not isinstance(limit, int) or isinstance(limit, bool) or limit < 1:
+            raise ValueError("sync executor admission limit must be a positive integer")
```
KonstantinMirin, line 95: This guard never executes under the suite, so `timed_sync_get_products_limit=0` producing a permanently saturated server at boot is unasserted. Add `SyncExecutorAdmission(0)` / `(-1)` / `(True)` raising `ValueError`.

While here: `release()` is unbounded — making it release twice (an ever-growing ceiling after every completed worker) leaves `tests/test_time_budget.py` and `tests/test_decisioning_dispatch.py` green at 84 passed. The assertion that pins release-exactly-once is: after N timed calls have fully completed, a burst of N+1 must still block exactly one.

KonstantinMirin (commented):
## Review — PR #1000

**Overview** — The direct-sync path now does what the title says. `asyncio.shield` over the wrapped worker plus `_settle_cancelled_sync_lifecycle` fires the lifecycle hooks from the worker's real terminal outcome, and `test_sync_create_media_buy_cancellation_waits_for_worker_success` proves the CONSUMING→CONSUMED transition survives a buyer disconnect. Permit accounting is single-release on both submit sites. Two things need work before this lands: the new `except BaseException` arm in `_run()` releases a proposal reservation when a background handoff task is cancelled, which lets a retry double-book a media buy; and both supervision gates key on `on_failure`, which `get_products` never wires, so the timed path the PR is named for is the one path that stays unsupervised.

### Should fix

Findings 2, 4, 5 and 10 are one root, already open as #1004: a policy re-detected by inspecting objects (`on_failure is not None`, `executor._max_workers`) or hand-copied to a subset of its sites, instead of owned at one boundary. Evidence from this PR is added there; the sites below stand on their own.

**1. Cancelling a background handoff task releases the reservation while the adopter's work is still outstanding**

`src/adcp/decisioning/dispatch.py:1972-1977`:

```python
except BaseException as exc:
    if on_failure is not None:
        await _safe_on_failure_call(on_failure, exc, method_name)
    raise
```

For `create_media_buy` that hook is `_release_reservation_hook` → `release_proposal_reservation` → CONSUMING → COMMITTED. Background handoff tasks are detached `create_task`s (`dispatch.py:2069`); ordinary loop/SIGTERM shutdown cancels them. Cancel one after the handoff fn has issued its upstream create:

```
HEAD   state after bg cancel: ProposalState.COMMITTED | registry row: submitted
       RETRY ACCEPTED -> {'media_buy_id': 'mb_duplicate_2'}
BASE   state after bg cancel: ProposalState.CONSUMING
       RETRY REJECTED -> AdcpError[PROPOSAL_NOT_COMMITTED / correctable]
```

One committed proposal, two media buys upstream. The arm also skips `_fail()`, so the registry row stays `submitted` with no terminal webhook — the buyer is left holding a task that never terminates, which is exactly what provokes the retry.

Root cause: this arm infers "our asyncio task was cancelled" ⇒ "the adopter's work did not happen". That is the inference the rest of this PR exists to refute on the request path. The handoff fn is not shielded here, and for a sync fn it is a `run_in_executor` thread that cannot be stopped at all.

Either do not fire `on_failure` from this arm (leave CONSUMING, which fails closed and is what eviction exists for), or mirror `_settle_cancelled_sync_lifecycle` — shield the fn and settle from its real terminal outcome. Either way a release must be paired with a terminal registry state so the buyer is not invited to double-book. Lines 1972-1977 never execute under the suite; add a test that cancels a background handoff task with a reserved proposal.

Separately, "then remain cancellation to the task scheduler" in that comment does not parse.

**2. Supervision keys on `on_failure`, so no timed `get_products` is supervised**

Raised in the aao-ipr-bot review of 2026-07-29 ("Routed, no-deadline `get_products` cancel drops its `on_complete`"), still open on this push, and wider than the routed no-deadline case — it covers the whole timed path.

Both settle sites gate on the same field:

- `src/adcp/decisioning/dispatch.py:1469` — `except asyncio.CancelledError: if on_failure is not None:`
- `src/adcp/decisioning/dispatch.py:1605` — `if isinstance(nested_sync_future, asyncio.Future) and on_failure is not None:`

`get_products` is the only method that carries a deadline and the only caller that passes `sync_admission`, and it wires `on_complete=_persist_draft_hook` with no `on_failure` (`src/adcp/decisioning/handler.py:1974-1979`). Cancel a sync `get_products` mid-worker and let the worker finish:

```
on_complete only     -> ON_COMPLETE CALLS: []
add on_failure=noop  -> ON_COMPLETE CALLS: [{'products': []}]
```

The gate is the whole difference. Three consequences follow from it:

- The persist-draft hook is silently dropped on every timed-out sync `get_products`.
- The shielded future has no owner in the else-branch, so a worker that raises after the deadline lands in asyncio's default handler. New at HEAD, absent at `b7ef1dfc`:
  ```
  ERROR:asyncio:Future exception was never retrieved
  future: <Future finished exception=RuntimeError('upstream token sk-SECRET-LEAK rejected')>
  Traceback (most recent call last): ... (full adopter traceback)
  ```
  That is the adopter's raw exception text at ERROR, outside the framework's error funnel — while the same PR strips `str(exc)` from `details.caused_by` so a platform raising on secret material cannot leak it. The routed branch has the same shape via `asyncio.create_task` (`platform_router.py:137-139`) and produces "Task exception was never retrieved".
- `_adcp_sync_worker_future`, set at `platform_router.py:166`, has one consumer (`dispatch.py:1604`) behind the same gate, so on the `get_products` path the marker is written and never read.

The same root shows up one function over: the new `pre_handoff_reject` → `on_failure` branch at `dispatch.py:1668-1669` is dead by construction. `pre_handoff_reject=` is passed by exactly two callers (`handler.py:1976` `get_products`, `handler.py:2676` `get_signals`); `on_failure=` by exactly one (`handler.py:2192` `create_media_buy`). No caller passes both, and line 1669 never executes.

Root cause: the lifecycle contract is the pair `(on_complete, on_failure)` and the supervision predicate reads one half of it. It should ask whether a lifecycle exists at all. The spawn block — `create_task(_settle_cancelled_sync_lifecycle(...))` + `_SUPERVISED_SYNC_LIFECYCLES.add` + `discard` — is also written out twice verbatim at `1471-1487` and `1607-1623`, which is what let both copies share the wrong predicate.

Gate both sites on `on_complete is not None or on_failure is not None`, extract the spawn into one `_supervise_sync_lifecycle(...)`, and attach a consumer to the shielded future in every branch so a post-cancellation worker failure is logged by the framework rather than by asyncio's GC hook. Add a cancellation test for an `on_complete`-only timed sync `get_products`, covering both the late-success and late-raise legs.

**3. `except BaseException` around awaited cleanup swallows the cancellation**

Three handlers widened from `except Exception` to `except BaseException` around an `await`, then log and continue:

- `src/adcp/decisioning/dispatch.py:1774` (`_safe_on_failure_call`)
- `src/adcp/decisioning/proposal_dispatch.py:755`
- `src/adcp/decisioning/proposal_dispatch.py:888`

`on_failure` hooks do durable-store I/O, so they contain await points where a cancellation lands. When it does, the hook's `CancelledError` is logged and discarded and the caller re-raises the original exception. Adopter raises `AdcpError`, request task is cancelled while `on_failure` is mid store round-trip:

```
HEAD   RESULT: cancellation SWALLOWED, task raised AdcpError[INTERNAL_ERROR / terminal]
       cancelled() -> False | reservation released -> []
BASE   RESULT: task honoured cancellation
       cancelled() -> True  | reservation released -> []
```

Worst of both: a graceful-shutdown drain doing `task.cancel(); await task` gets a completed task back, and the reservation release is left half-done anyway.

Root cause: log-and-continue is the right policy for `Exception` and the wrong one for a `BaseException` that is not an `Exception`. Catch `BaseException` so the cleanup runs, then re-raise when the caught exception is not an `Exception` — the hook's cancellation is not the framework's to absorb.

One more at `proposal_dispatch.py:747`: the widened handler guards the `try:` at lines 698-746, and that block contains no `await` — `_derive_packages`, the `params.packages` assignment and `validate_capability_overlap` are all synchronous. `CancelledError` cannot be delivered there, so the new comment ("Cancellation and shutdown must not strand a durable reservation") describes a state the code cannot reach, and no test can be written for it. Narrow it back, or state the reachable trigger (a synchronous `SystemExit`/`KeyboardInterrupt` inside derivation) and test that. Lines 755, 756 and 888 never execute under the suite.

**4. The admission limit is read off `executor._max_workers`**

`src/adcp/decisioning/handler.py:1311-1317`:

```python
worker_count = int(getattr(executor, "_max_workers", 1))
admission_limit = ... else max(1, worker_count // 2)
```

`serve.py:120-124` documents BYO executors as "for operators with audit-instrumented thread pools or **wrappers around stdlib's executor**". A wrapper is exactly the shape that has no `_max_workers`:

```
wrapper around a 64-worker pool -> admission limit 1
plain 64-worker pool            -> admission limit 32
```

Every deadline-managed sync `get_products` in that deployment then serialises behind one permit held for the full worker duration, with no warning and no public way to observe the value. The parameter is annotated `executor: ThreadPoolExecutor` (`handler.py:1275-1278`), so by the declared type the `getattr` default is dead code; for the shape the docs advertise, it drops the limit to 1 with no warning. `int()` on an executor whose `_max_workers` is not int-able raises at handler construction.

Root cause: the composition root owns pool sizing — `serve.py` already resolved `thread_pool_size` / `_default_thread_pool_size()` at `serve.py:70,268` — and the handler re-derives it downstream from a CPython-internal field that is not part of any stability contract. Thread the resolved worker count (or the resolved admission limit) from `create_adcp_server_from_platform` into `PlatformHandler` and keep `timed_sync_get_products_limit` as the explicit override. If a BYO executor cannot supply a count, require the override rather than silently choosing 1.

**5. Supervised submit has no owned home: written twice, and a third site was left behind**

`src/adcp/decisioning/dispatch.py:1445-1467` and `src/adcp/decisioning/platform_router.py:141-162` run the same sequence in the same order — `await admission.acquire()` → `executor.submit(call)` → `except BaseException: admission.release(); raise` → a `_release_admission` done-callback wrapping `loop.call_soon_threadsafe(admission.release)` in `try/except RuntimeError` → `asyncio.wrap_future(..., loop=loop)` → `await asyncio.shield(worker)`. They differ only in identifiers and indentation, and already diverge on the fallback branch and on what happens at cancellation. Every step is load-bearing for permit accounting.

The third framework sync-submit site was not migrated. `src/adcp/decisioning/proposal_dispatch.py:244` is still the pre-PR shape:

```python
result = await loop.run_in_executor(
    executor, functools.partial(ctx_snapshot.run, method, finalize_req, ctx)
)
```

No shield, no admission, no settlement. Cancel a sync `ProposalManager.finalize_proposal` and the asyncio side unwinds immediately while the thread runs to completion and applies the adopter's finalize side effects; the framework's `store.commit` below never runs:

```
worker completed: True
proposal state after cancelled finalize: ProposalState.DRAFT
```

That is the same "worker mutated state, ledger disagrees" failure the PR body claims to close, on the file this PR opened to close it.

Root cause: "submit sync adopter work under supervision" is inline in `dispatch` and re-implemented in `platform_router`, so the third call site has nothing to call. `SyncExecutorAdmission` (`time_budget.py:82`) exposes only `acquire`/`release` and leaves the ordering contract — release exactly once, from the concurrent future's callback, marshalled back to the loop — to be re-derived at each site. Extract one `submit_supervised(executor, admission, call) -> asyncio.Future` next to the semaphore and route all three callers through it, with an e2e test for the cancelled finalize alongside `test_sync_create_media_buy_cancellation_waits_for_worker_success`.

**6. The settle task reruns the full projection, so a cancelled request can mint a task the buyer never learns about**

`_settle_cancelled_sync_lifecycle` (`src/adcp/decisioning/dispatch.py:1715`) calls `_project_invocation_result` at `dispatch.py:1737`, which reruns the full result projection rather than only the lifecycle hooks. When the cancelled sync adopter returns a `TaskHandoff`, the settle path runs `_project_handoff`: it issues a registry `task_id`, launches the background handoff, persists the terminal artifact, and emits the completion webhook if a push config was supplied. The `Submitted` envelope it builds is then discarded — there is no waiter. Cancel a sync `create_media_buy` that returns a handoff, then release the worker:

```
REGISTRY RECORDS: {'task_dc9d25f8c16e4add': TaskRecord(state='completed',
  task_type='create_media_buy', result={'media_buy_id': 'mb_orphan', ...})}
```

Absent a push config the buyer holds no `task_id`, so the task is unreachable via `tasks/get` and their only recourse is a retry that re-executes the work. This arm could not run before the PR (`await run_in_executor` raised `CancelledError` before the handoff branch), so the diff introduces it, and neither of the two new dispatch tests exercises it — they cover a plain dict return and a `RuntimeError`.

Decide and state the contract: either restrict the settle path to the lifecycle hooks it exists for and refuse to promote a handoff once the caller is gone, or keep the promotion and make the task recoverable — log the issued `task_id` with the request correlation id at WARNING, and name the push-notification surface (`schemas/cache/3.1/core/protocol-envelope.json` @ AdCP 3.1.8) as the delivery channel that makes it legal. Either way pin the chosen behavior with a test for the `TaskHandoff` and `WorkflowHandoff` arms.

**7. The bound executor is ignored without an admission controller, and the worker future travels on the exception**

Two defects, both in `_run_sync_delegate`.

`src/adcp/decisioning/platform_router.py:137` reads `if admission is None or executor is None:` → `asyncio.to_thread(...)`. `_bind_routed_sync_execution(sync_admission, executor)` always binds a non-`None` executor (`dispatch.py:1425`), but the delegate only uses it when an admission controller is also present, i.e. only for deadline-managed `get_products`. Every other routed sync child — `create_media_buy`, `update_media_buy`, `refine_get_products`, all synthesized delegates — runs on the loop's default executor, outside the adopter's BYO pool, with the right executor sitting in the ContextVar one line above. With a BYO pool named `FRAMEWORK`, a routed sync child on a no-deadline `get_products` ran on thread `asyncio_0`. That undercuts the D5 BYO-executor contract documented in `serve.py`.

`src/adcp/decisioning/platform_router.py:166` signals dispatch by `setattr(exc, "_adcp_sync_worker_future", worker)`, read back at `dispatch.py:1604` as `Any` and narrowed only by `isinstance(..., asyncio.Future)`. `platform_router` is a `DecisioningPlatform` implementation below dispatch, and it now encodes dispatch's settlement protocol on an exception instance. Nothing fails at type-check time if the attribute name, producer or consumer drifts, and the channel is lossy: any frame between the router and dispatch that catches `CancelledError` and re-raises a fresh instance (adopter middleware, a wrapping platform) drops the marker with no log at either end.

Root cause for both: the executor choice is execution plumbing, not time-budget policy, and putting the `ThreadPoolExecutor` ContextVar in `time_budget.py` next to the deadline machinery is what made "no deadline" read as "no configured executor" and left the live worker future with nowhere typed to live. Split the condition (`executor is not None` → submit to it; `admission is not None` → additionally gate on a permit), carry the worker future on the same scope object the two modules already share (or on a `SyncWorkerCancelled(CancelledError)` subclass with a typed `worker: asyncio.Future[Any]`), and log when dispatch unwinds a cancelled routed sync call with no marker. Add a test asserting a routed sync child runs on the handler's executor (`thread_name_prefix`) with and without a time budget.

**8. New surfaces shipped without a test that fails without them**

Four groups, each verified by mutation or by coverage of the added lines:

- **The public knob.** `timed_sync_get_products_limit` was added to `create_adcp_server_from_platform` (`serve.py:84,381`) and `serve` (`serve.py:466,588`), and appears in tests only as a direct `PlatformHandler(...)` kwarg (`tests/test_time_budget.py:312,393,447`). Deleting both plumbing lines leaves the suite green — the parameter becomes a documented no-op and nothing notices. The `ValueError` guard at `time_budget.py:93` never executes, so `timed_sync_get_products_limit=0` producing a permanently saturated server at boot is unasserted, as is the documented `max(1, worker_count // 2)` default.
- **Permit accounting on return and on submit failure.** Making `SyncExecutorAdmission.release()` release twice — an unbounded ceiling after every completed worker — leaves `tests/test_time_budget.py` and `tests/test_decisioning_dispatch.py` green (84 passed). The submit-failure release paths at `dispatch.py:1449-1452` and `platform_router.py:148-150` never execute. Add: an executor stub whose `submit` raises `RuntimeError` (post-`shutdown` behavior) → the next timed call is still admitted; and after N timed calls fully complete, a burst of N+1 blocks exactly one.
- **The lazy proposal-manager arm.** `platform_router.py:1099` is the only one of the eight migrated `asyncio.to_thread` → `_run_sync_delegate` sites with no execution. `test_router_sync_timeout_uses_bounded_admission[lazy]` covers the platform arm, not `proposal_manager_for_tenant`, so that path silently changed thread-dispatch semantics with no test.
- **The settle task's failure arms.** `dispatch.py:1750,1754` (`except BaseException` + `logger.exception`) and `dispatch.py:1774` never execute; reverting `_safe_on_failure_call` to `except Exception` leaves the suite green.

**9. Admission saturation emits the timeout `incomplete[]` verbatim**

The bounded-admission path introduces a genuinely new server-side condition — the seller never searched, because no permit came free before the budget expired — and reuses the pre-existing timeout payload word for word (`src/adcp/decisioning/time_budget.py:219-233`): `"time_budget exhausted (N unit); return the best results achievable within the budget. Retry with a larger time_budget…"`. Per `schemas/cache/3.1/media-buy/get-products-response.json` @ AdCP 3.1.8, `incomplete[]` "Declares what the seller could not finish within the buyer's time_budget **or due to internal limits**", and the per-entry `description` is the "Human-readable explanation of what is missing and why". The spec separates exactly these two causes; today a buyer cannot tell "searched and ran out of time" from "declined admission and searched nothing". `scope: "products"` is correct on both.

Grading is thin on the same path: both new saturation tests assert only `getattr(result, "incomplete", None)` truthiness, so a regression emitting an off-enum `scope` or an empty `incomplete[]` (`minItems: 1`) would still pass. Give the saturation path its own `description` naming the internal admission limit, and assert the full projected `incomplete[0]` (`scope`, `description`, `products == []`).

**10. The sanitized `caused_by` shape was made uniform, then stopped one site short**

The diff imports `_internal_error_details` into `handler.py` and routes the `get_adcp_capabilities` INTERNAL_ERROR through it (`handler.py:1650-1653`), dropping `str(exc)` from `details.caused_by`. That is the right call and the new assertions at `tests/test_decisioning_capabilities_projection.py:640-642` pin it. Three follow-throughs are missing:

- `handler.py:2107` still hand-rolls `details={"caused_by": {"type": type(exc).__name__}}` for the `ProductConfigStore.lookup_implementation_configs` `SERVICE_UNAVAILABLE` wrap — the same wire shape under a different name, so the next change to the sanitized `details` contract has to be made twice. Route it through `_internal_error_details(exc)`, deciding explicitly whether the helper's `details.validation_errors` addition is wanted there, or factor the `caused_by`-only core into a helper both call.
- `dispatch.py:1583-1586` still reads "We expose only the exception class name + str (not the traceback)" directly above the `_internal_error_details(exc)` call, which emits `{"type": ...}` only. This PR makes that sanitization the single shape, so the comment is now the one place telling a future maintainer to put `str(exc)` back on the wire. Drop "+ str" and state the invariant positively: class name only, message in the server log via `logger.exception`. Raised in the 2026-07-29 review, still open.
- The change is buyer-visible on two values — the `message` string and the loss of `details.caused_by.message` — and reaches MCP and A2A buyers. The PR's Compatibility section says only "Existing direct asynchronous platform behavior is unchanged". Note the `details.caused_by.message` removal there and in the changelog, and cite `schemas/cache/3.1/core/error.json` @ AdCP 3.1.8 (`details` open with `additionalProperties: true`, `caused_by` non-normative, `recovery` unchanged at `terminal`) as the grounding.

### Notes

- Permit accounting under a cancelled `acquire()` was the shape most likely to bite — short `time_budget` while permits are held, each `wait_for` cancelling a pending `acquire`, capacity ratcheting to zero. 200 cancelled acquires per trial, 5 trials, on 3.12 and on 3.11's `asyncio.Semaphore`: effective capacity returned to exactly the limit every time. Not exercised on 3.10, which ships the older `Semaphore.acquire` — worth one CI run.
- Per-tenant fairness of `SyncExecutorAdmission` is out of scope: one tenant can hold every permit, but the diff replaces no bound with a bound, and keying the semaphore on `ctx.account` is a capacity-policy decision rather than a defect this PR introduces.
- `_SUPERVISED_SYNC_LIFECYCLES` is a module-global task set with no shutdown drain. Out of scope: it follows the module's existing convention (`_BACKGROUND_HANDOFF_TASKS` at `dispatch.py:2091`, `_BACKGROUND_WEBHOOK_TASKS` at `webhook_emit.py:107`), so changing it is a repo-wide decision.
- `tests/test_pg_idempotency_backend.py::test_delete_expired_defaults_to_wall_clock` failed once across 16 full-suite runs, while two suites ran concurrently, and passed in the other 15. Pre-existing wall-clock sensitivity, unrelated to this diff — recorded so the single red line in the logs is not read as a finding.

src/adcp/decisioning/handler.py
```diff
@@ -1630,17 +1647,9 @@ async def get_adcp_capabilities(
             )
             raise AdcpError(
                 "INTERNAL_ERROR",
-                message=(
-                    "Unhandled exception in platform.get_adcp_capabilities_for_request: "
-                    f"{type(exc).__name__}: {exc}"
-                ),
+                message=_internal_error_message("get_adcp_capabilities_for_request", exc),
```
KonstantinMirin, line 1671: Routing this through `_internal_error_message` / `_internal_error_details` is the right call — it drops `str(exc)` so a platform raising on secret material cannot leak it on the wire, and the new assertions at `tests/test_decisioning_capabilities_projection.py:640-642` pin that. Three follow-throughs are missing:

1. `handler.py:2107` still hand-rolls `details={"caused_by": {"type": type(exc).__name__}}` for the `ProductConfigStore.lookup_implementation_configs` `SERVICE_UNAVAILABLE` wrap — same wire shape under a different name, so the next change to the sanitized `details` contract has to be made twice. Decide explicitly whether the helper's `details.validation_errors` addition is wanted there, or factor the `caused_by`-only core into a helper both call.
2. `dispatch.py:1583-1586` still reads "We expose only the exception class name + str (not the traceback)" directly above the `_internal_error_details(exc)` call, which emits `{"type": ...}` only. Drop "+ str" — as written it is the one place telling a future maintainer to put the leak back. Raised in the 2026-07-29 review, still open.
3. This is buyer-visible on two values (the `message` string, and the loss of `details.caused_by.message`) and reaches MCP and A2A buyers. The Compatibility section says only "Existing direct asynchronous platform behavior is unchanged". Note the `details.caused_by.message` removal there and in the changelog, citing `schemas/cache/3.1/core/error.json` @ AdCP 3.1.8 (`details` open with `additionalProperties: true`, `caused_by` non-normative, `recovery` unchanged at `terminal`).

bokelley:
Superseded by #1025, which incorporates this work and resolves the remaining cancellation concern by keeping async mutation reservations fail-closed on cancellation.

